PR #6, opened 2024-06-19
MLT-0016 Create Docker Compose file

anotheroneofthese:
The compose file is planned to be used for local deployment and testing of Hermes. 
For now this only includes a local deployment of MongoDB + Mongo Express, such that we can start developing how the database itself looks.


docker/compose.yaml
```diff
@@ -1,3 +1,19 @@
...
+  mongo-express:
+    image: mongo-express
+    restart: always
+    ports:
+      - 8081:8081
+    depends_on:
+      - mongodb
+    environment:
+      ME_CONFIG_MONGODB_ADMINUSERNAME: root
+      ME_CONFIG_MONGODB_ADMINPASSWORD: example
```
MormonJesus69420, line 17: Not using toor as root pass, for shame... 😔

MormonJesus69420 (approved):
Approved, great job, did some small changes to the way keycloak gets spun up and the realm config. Otherwise it was a great start

PR #9, opened 2024-07-02
MLT-0017 Product Data Structure

anotheroneofthese:
This PR adds data structures to accomodate the various products we will be storing in WLS


src/main/kotlin/no/nb/mlt/wls/core/data/Packaging.kt
```diff
@@ -0,0 +1,8 @@
+package no.nb.mlt.wls.core.data
+
+enum class Packaging {
+    OBJ,
+    ESK,
+    ABOX,
+    NONE
```
MormonJesus69420, line 7: Need to decide if this should be also a DTO, since in SynQ and API we use different terms. So we should probably have enums for both DTOs and internal logic.
In API we have BOX, ABOX, CRATE, and NONE. 
In SynQ we have ESK, ABOX, EA (Each, I think that one's not actually used in production), and OBJ

src/main/kotlin/no/nb/mlt/wls/product/dto/ProductDTO.kt
```diff
@@ -0,0 +1,20 @@
+package no.nb.mlt.wls.product.dto
+
+import io.swagger.v3.oas.annotations.media.Schema
+import no.nb.mlt.wls.core.data.HostName
+import no.nb.mlt.wls.core.data.Owner
+import no.nb.mlt.wls.core.data.Packaging
+import no.nb.mlt.wls.core.data.PreferredEnvironment
+
+@JvmRecord
```
MormonJesus69420, line 9: I think this can be removed

MormonJesus69420 (changes requested):

src/main/kotlin/no/nb/mlt/wls/product/repository/ProductsRepository.kt
```diff
@@ -0,0 +1,9 @@
+package no.nb.mlt.wls.product.repository
+
+import no.nb.mlt.wls.core.data.HostName
+import no.nb.mlt.wls.product.model.ProductModel
+import org.springframework.data.mongodb.repository.MongoRepository
+
+interface ProductsRepository : MongoRepository<ProductModel, String> {
```
MormonJesus69420, line 7:
```suggestion
interface ProductRepository : MongoRepository<ProductModel, String> {
```

src/main/kotlin/no/nb/mlt/wls/product/service/ProductsService.kt
```diff
@@ -0,0 +1,17 @@
+package no.nb.mlt.wls.product.service
+
+import no.nb.mlt.wls.core.data.HostName
+import no.nb.mlt.wls.product.model.ProductModel
+import no.nb.mlt.wls.product.repository.ProductsRepository
+import org.springframework.stereotype.Service
+
+@Service
+class ProductsService(val db: ProductsRepository) {
```
MormonJesus69420, line 9:
```suggestion
class ProductService(val db: ProductsRepository) {
```

MormonJesus69420 (changes requested):
Forgot about those two

src/main/kotlin/no/nb/mlt/wls/core/data/Packaging.kt
anotheroneofthese: Added a PackagingDTO for this purpose

src/main/kotlin/no/nb/mlt/wls/core/config/SecurityConfig.kt
```diff
@@ -53,6 +53,7 @@ class SecurityConfig {
                     ).permitAll()
                     .pathMatchers(HttpMethod.GET, "/v1/test/authorized").hasAuthority("wls-role")
                     .pathMatchers(HttpMethod.GET, "/v1/test/authenticated").authenticated()
+                    .pathMatchers("/v1/test/json").authenticated()
```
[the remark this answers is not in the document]
anotheroneofthese: Will write a todo on this

MormonJesus69420 (approved):
Looks good after changes suggested by Tom

PR #10, opened 2024-07-05
MLT-0023 Update secrets with correct names

anotheroneofthese:
Fixes the deployment being broken due to looking for the wrong secrets


MormonJesus69420 (approved):
Looks great to me

PR #11, opened 2024-07-05
MLT-0012 Add products to SynQ

anotheroneofthese:
This PR implements the functionality to connect and add products to our SynQ storage system.


src/main/kotlin/no/nb/mlt/wls/core/config/SecurityConfig.kt
MormonJesus69420, line 1: Shouldn't we change to per-path authorization? Might also consider making everything require authentication, and open as needed. GET methods may be authentication only, but everything else should require authorization with specific roles.  
Need to decide what roles we need, and then add them to both local and stage Keycloak.
Swagger may be open for everyone.

src/main/kotlin/no/nb/mlt/wls/product/controller/ProductController.kt
MormonJesus69420, line 1: Good for now, might want to add OpenAPI annotations so we give more help/info to users

src/main/kotlin/no/nb/mlt/wls/product/payloads/ApiProductPayload.kt
```diff
@@ -23,7 +23,7 @@ data class ApiProductPayload(
     @Schema(example = "1.0")
     val quantity: Float,
     @Schema(example = "NONE")
-    val preferredEnvironment: PreferredEnvironment,
+    val environment: Environment,
```
MormonJesus69420, line 26: This should still be called preferredEnvironment in the outward facing API, we can just call the enum Environment for ease of use and since we will likely re-use it when we start storing information about preferred env and actual storage env.  
Sorry for not making myself clear here.

src/main/kotlin/no/nb/mlt/wls/product/payloads/SynqProductPayload.kt
```diff
@@ -1,24 +1,86 @@
...
+        Owner.NB -> NB
+    }
+}
+
+fun Packaging.toSynqPackaging(): SynqPackaging {
+    return when (this) {
+        Packaging.NONE -> OBJ
+        Packaging.BOX -> ESK
+    }
 }
```
MormonJesus69420, line 86: Well done, although I think those can be simplified a bit in Kotlin to an expression body, for example:
```kotlin
fun SynqPackaging.toPackaging(): Packaging = when (this) {
    OBJ -> Packaging.NONE
    ESK -> Packaging.BOX
}
```

src/main/kotlin/no/nb/mlt/wls/product/repository/ProductRepository.kt
```diff
@@ -6,4 +6,6 @@ import org.springframework.data.mongodb.repository.MongoRepository
 
 interface ProductRepository : MongoRepository<Product, String> {
     fun findByHostName(hostName: HostName): List<Product>
+
+    fun existsByHostId(id: String): Boolean
```
MormonJesus69420, line 10: Now that I think about it, I think we need to check existence both by host ID and name, since we can have duplicate IDs across different hosts

src/main/kotlin/no/nb/mlt/wls/product/payloads/SynqProductPayload.kt
```diff
@@ -1,24 +1,86 @@
...
         NB
     }
 
     class Barcode(val barcodeId: String)
 
-    class ProductUom(val uomId: Packaging)
+    class ProductUom(val uomId: SynqPackaging)
+}
+
+// TODO - Will we ever receive this payload?
```
MormonJesus69420, line 37: Probably no, can either leave it for consistency's sake, or delete it for simplicity's sake

src/main/kotlin/no/nb/mlt/wls/product/service/SynqService.kt
```diff
@@ -0,0 +1,24 @@
...
+@Service
+class SynqService {
+    val restTemplate: RestTemplate = RestTemplate()
+
+    @Value("\${synq.path.base}")
+    lateinit var baseUrl: String
+
+    fun createProduct(payload: SynqProductPayload) {
+        // TODO - Validation
+        val uri = URI.create("$baseUrl/nbproducts")
```
MormonJesus69420, line 20: Note to self: check if there are safer ways of creating URIs

MormonJesus69420 (changes requested):
Looks great, small stuff that we need to discuss some more and work on improving

src/main/kotlin/no/nb/mlt/wls/core/config/SecurityConfig.kt
anotheroneofthese: As we went through today, we will keep this in SecurityConfig for now

src/main/resources/application.yml
```diff
@@ -60,13 +61,20 @@ spring:
       resourceserver:
         jwt:
           issuer-uri: http://localhost:8082/realms/mlt-local
+          audiences: http://localhost:8080/hermes
```
anotheroneofthese, line 64: Is this used? @MormonJesus69420

MormonJesus69420: Yes, that's what validates the JWT audience now, instead of doing it by hand

src/main/kotlin/no/nb/mlt/wls/product/service/SynqService.kt
MormonJesus69420: It seems to be possible to use the UriComponentsBuilder from Spring Framework, however, in our simple case it seems like an overkill for a simple problem
```kotlin
val uri = UriComponentsBuilder
    .fromHttpUrl(baseUrl)
    .path("/nbproducts")
    .build(emptyMap<String, String>())
```

MormonJesus69420 (approved):
I think that's the best we can get done before testing it in stage with Axiell

PR #13, opened 2024-07-15
Hotfix fix ingress spec in deployment

MormonJesus69420:
Fixed the ingress spec for k8s deployment, added graceful shutdown to spring server


anotheroneofthese (approved):
Surely it will work this time

MormonJesus69420:
> Surely it will work this time

I'll try spinning that's a good trick

PR #16, opened 2024-07-18
MLT-0024 Add initial Product Controller tests

anotheroneofthese:
Tests the endpoint for creating new products using mocking


src/test/kotlin/no/nb/mlt/wls/HermesApplicationTests.kt
MormonJesus69420, line 1: This test class can probably be removed, since it doesn't really test much, unless we want to have 100% test coverage. At the very least, we should get right of `itsFunToNOTFail` method

src/test/kotlin/no/nb/mlt/wls/controller/ProductControllerTest.kt
```diff
@@ -0,0 +1,49 @@
...
+        )
+
+    @Test
+    fun saveProductTest() {
+        every {
+            productService.save(testProduct.toApiPayload())
+        } returns ResponseEntity(testProduct.toApiPayload(), HttpStatus.OK)
+
+        val result = productService.save(testProduct.toApiPayload())
+        assertEquals(HttpStatus.OK, result.statusCode)
```
MormonJesus69420, line 47: This doesn't really test anything, you just test that the mock returns what you told it to return. Should test using `@WebMvcTest` and do actual HTTP calls to endpoint.
Might also want to use the actual service and database to ensure that we have that part of the code working. We should probably mock the SynQ service to avoid calling SynQ or just mock response from `RestTemplate` inside the `SynqService` class (will have to use Autowire to inject the mocked `RestTemplate` or some other form of DI)
That way we can test much more cases and ensure that it's all handled correctly (see return codes from method in Swagger for test cases), along with Client Auth

MormonJesus69420 (changes requested):

src/test/kotlin/no/nb/mlt/wls/controller/ProductControllerTest.kt
MormonJesus69420: Sorry it should be `@WebFluxTest`, not `@WebMvcTest`

anotheroneofthese:
Reworked the tests to test everything on a controller level. The local TestContainer running MongoDB is now actually being used to test out the full flow. Only the SynqController remains, as the mocking of it currently fails (this is not present in the unit test, since the SynqController is never queried due to the product already existing)

src/main/kotlin/no/nb/mlt/wls/product/payloads/ApiProductPayload.kt
```diff
@@ -7,7 +7,6 @@ import no.nb.mlt.wls.core.data.Owner
 import no.nb.mlt.wls.core.data.Packaging
 import no.nb.mlt.wls.product.model.Product
 
-// TODO - Enforce style with validation. Should valid fields be returned in an error message, or be declared in the description?
 @Schema(
     description = "Payload for registering a product in Hermes WLS, and appropriate storage system for the product.",
     example = """
```
MormonJesus69420, line 12: Just a small thing, we should remove location and quantity from the example, I goofed up when I wrote it

src/test/kotlin/no/nb/mlt/wls/controller/ProductControllerTest.kt
```diff
@@ -0,0 +1,85 @@
...
+
+    val testProduct =
+        Product(
+            hostName = HostName.AXIELL,
+            hostId = "mlt-2048",
+            productCategory = "BOOK",
+            description = "Ringenes Herre samling",
+            packaging = Packaging.BOX,
+            location = "SYNQ_WAREHOUSE",
+            quantity = 1.0,
```
MormonJesus69420, line 45: Quantity = 0 is a more appropriate value. We should consider either replacing whatever value we get for quantity and location in requests or throwing an error.

MormonJesus69420 (changes requested):

MormonJesus69420 (approved):
LG2M

PR #18, opened 2024-07-23
MLT-0028 fix error in product creation

MormonJesus69420:
Change how we handle the location and quantity in the product endpoint. We used to receive them as is and forward them to the database and other places. That does not make sense as the catalogue does not decide the quantity nor the location. Now the endpoint ignores them and overrides these fields with default values.
The swagger doc reflects the change by making those two fields read only.


anotheroneofthese:
Should `location` be annotated with `@Nullable` to reflect this?

anotheroneofthese:
Actually, scratch that. We are using Kotlin after all

anotheroneofthese (approved):

PR #25, opened 2024-08-08
MLT-0034 Reactive migration

anotheroneofthese:
This PR attempts to migrate parts of the service to use the Reactive paradigm. The idea being to squeeze some extra performance out of our service


src/main/kotlin/no/nb/mlt/wls/order/controller/OrderController.kt
```diff
@@ -33,7 +33,8 @@ class OrderController(val orderService: OrderService) {
             responseCode = "400",
             description =
                 """Order payload is invalid and was not created.
-                Error message contains information about the invalid fields.""",
+                An empty error message means the order already exists with the current ID.
+                Otherwise, the error message contains information about the invalid fields.""",
```
MormonJesus69420, line 37: Need to discuss this behaviour with others, should we handle it the same way we handle products? Meaning if someone tries to create an order that already exists we return a 200 with existing order info. I think this would be more consistent.

src/main/kotlin/no/nb/mlt/wls/order/controller/OrderController.kt
```diff
@@ -48,7 +49,7 @@ class OrderController(val orderService: OrderService) {
         )
     )
     @PostMapping("/order/batch/create")
-    fun createOrder(
+    suspend fun createOrder(
         @RequestBody payload: ApiOrderPayload
     ): ResponseEntity<ApiOrderPayload> = orderService.createOrder(payload)
```
MormonJesus69420, line 54: Correct me if I'm wrong should the return type be changed here to something more reactive, or do we need to return ResponseEntity

src/main/kotlin/no/nb/mlt/wls/product/service/SynqProductService.kt
```diff
@@ -3,27 +3,36 @@ package no.nb.mlt.wls.product.service
...
+import org.springframework.web.reactive.function.BodyInserters
+import org.springframework.web.reactive.function.client.WebClient
+import org.springframework.web.reactive.function.client.awaitBody
 import org.springframework.web.server.ServerErrorException
 import java.net.URI
 
 @Service
 class SynqProductService {
-    val restTemplate: RestTemplate = RestTemplate()
+    val webClient: WebClient = WebClient.builder().build()
```
MormonJesus69420, line 18: Isn't it better to autowire it in, like in the SynqOrderService?

src/test/kotlin/no/nb/mlt/wls/product/controller/ProductControllerTest.kt
```diff
@@ -65,7 +69,7 @@ class ProductControllerTest(
     @WithMockUser
     fun `createProduct with valid payload creates product`() {
         every {
-            synqProductService.createProduct(any())
+            runBlocking { synqProductService.createProduct(any()) }
```
MormonJesus69420, line 72: Have you just tried calling block() instead? Also if we need more complex testing then I think it would be good to use either a step verifier especially when/if we work with fluxes:
https://www.baeldung.com/reactive-streams-step-verifier-test-publisher
Kotlin seems to have a nice extension for it:
https://stackoverflow.com/a/66213441

MormonJesus69420 (changes requested):

src/test/kotlin/no/nb/mlt/wls/product/controller/ProductControllerTest.kt
anotheroneofthese: Originally when writing this I was running into issues where making blocking calls wasn't allowed. Not sure exactly what the context was, but it seems like this is now possible. Will do that

anotheroneofthese: Double-checked why this was implemented this way and found the reasoning:
You aren't meant to block objects that return Reactive elements, and as such if you are on a thread you can get an error message like this: 
`java.lang.IllegalStateException: block()/blockFirst()/blockLast() are blocking, which is not supported in thread parallel-2`
This happens if you try blocking any `Mono` from a `ReactiveMongoRepository`, which makes sense. Using this makes Mongo send out their objects in a reactive manner, so blocking them will just make it act as a regular `MongoRepository` - entirely defeating the purpose to begin with. 
Because I ran into this issue first, I didn't consider blocking the SynQ services as I thought I would run into the same issue. This assumption is wrong, as SynQ by nature is a blocking application, and therefore we can block it.

src/main/kotlin/no/nb/mlt/wls/order/controller/OrderController.kt
anotheroneofthese: This depends on what we want regarding potential/future consumers of our own API. If we want to be able to build the service to act in a reactive manner, meaning sending partial data or streaming the data, then the answer is yes. Otherwise a `ResponseEntity` should be fine for now.

MormonJesus69420 (approved):
Good for now. We'll have to do some more tests in the future and make sure we are fulfilling the requirements from others. But it's a good start for now

anotheroneofthese:
Have tested it and it seems to work with the current local and staging environment

MormonJesus69420 (approved):

PR #26, opened 2024-08-08
MLT-0035 Dummy SynQ deployment

anotheroneofthese:
Adds the improved dummy-synq server to the Docker Compose deployment, and as a dependency for integration tests via Testcontainers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,14 +112,14 @@ docker pull harbor.nb.no/mlt/wls:<TAG>
 With the image either built or pulled, it can be run using the following command:
 
 ```shell
-docker run -p 8080:8080 -e SPRING_PROFILES_ACTIVE='dev' harbor.nb.no/mlt/wls:<TAG>
+docker run -p 8080:8080 -e SPRING_PROFILES_ACTIVE='local-dev' harbor.nb.no/mlt/wls:<TAG>
 ```
 
 ### Using an IDE
 
 For local development and testing an IDE like [IntelliJ IDEA](https://www.jetbrains.com/idea/) is recommended.
 Its default Spring run configuration for the application works well.
-To test the service with authentication make sure that dev version of the Keycloak is running, and set the `SPRING_PROFILES_ACTIVE` variable to `dev`.
+To test the service with authentication make sure that dev version of the Keycloak is running, and set the `SPRING_PROFILES_ACTIVE` variable to `local-dev`.
 In case you can't/won't use local Keycloak instance, then provide the `KEYCLOAK_ISSUER_URI` variable (see below) and set the `SPRING_PROFILES_ACTIVE` variable to `stage`.
 Keycloak in `dev` and `stage` is set up with a test client `wls`.
 
@@ -224,7 +224,7 @@ Although it's better to run them on your machine before pushing as it will save 
 The following environment variables are relevant to configuring the application:
 
 - `KEYCLOAK_ISSUER_URI`: Is used to point at the Keycloak server used for authentication (required)
-- `SPRING_PROFILES_ACTIVE`: Is used to set the active Spring profile, use `dev` or `stage` for testing authentication (optional, default is `pipeline`)
+- `SPRING_PROFILES_ACTIVE`: Is used to set the active Spring profile, use `local-dev` or `stage` for testing authentication (optional, default is `pipeline`)
 - `MONGODB_USERNAME`: Is the username used for connecting to MongoDB (required)
 - `MONGODB_PASSWORD`: Is the password used for connecting to MongoDB (required)
 - `MONGODB_DATABASE`: Is the name of the database that Hermes should use (required)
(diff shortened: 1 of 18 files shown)


anotheroneofthese:
How should we handle this in regards to the Docker image? Tests and local development will likely fail outside NLN's network.

src/main/kotlin/no/nb/mlt/wls/product/payloads/SynqProductPayload.kt
```diff
@@ -26,7 +26,7 @@ data class SynqProductPayload(
         ESK
     }
 
-    data class Barcode(val barcodeId: String)
+    data class Barcode(val barcodeId: String, val barcodeDesc: String)
```
MormonJesus69420, line 29: Think it's actually better to delete barcodeDescription in the dummy-synq. I forgot to do it with all the meetings going on. Since it's not required and we don't seem to have a use for it.

src/test/kotlin/no/nb/mlt/wls/TestcontainerInitializer.kt
```diff
@@ -3,22 +3,24 @@ package no.nb.mlt.wls
...
 import org.springframework.context.ConfigurableApplicationContext
+import org.testcontainers.containers.GenericContainer
 import org.testcontainers.containers.MongoDBContainer
 import org.testcontainers.lifecycle.Startables
 import org.testcontainers.utility.DockerImageName
 
 class TestcontainerInitializer : ApplicationContextInitializer<ConfigurableApplicationContext> {
     companion object {
         val MongoContainer = MongoDBContainer(DockerImageName.parse("mongo:4.4.22"))
+        val DummySynqContainer = GenericContainer(DockerImageName.parse("harbor.nb.no/mlt/dummy-synq:main"))
```
MormonJesus69420, line 14: To avoid failures outside of NLN network during test we might just declare a variable for Docker Image name/location so you can easily change to using a locally built one instead of pulling one from NLN's harbor instance.

src/test/kotlin/no/nb/mlt/wls/product/ProductConvertiontests.kt
```diff
@@ -45,7 +45,7 @@ class ProductConvertiontests {
         SynqProductPayload(
             productId = "mlt-test-1234",
             owner = SynqOwner.NB,
-            barcode = SynqProductPayload.Barcode("mlt-test-1234"),
+            barcode = SynqProductPayload.Barcode("mlt-test-1234", "barcode"),
```
MormonJesus69420, line 48: Same comment as in SynqProductPayload.kt

MormonJesus69420 (changes requested):

src/main/kotlin/no/nb/mlt/wls/product/payloads/SynqProductPayload.kt
anotheroneofthese: Will make a PR for this :+1:

src/test/kotlin/no/nb/mlt/wls/TestcontainerInitializer.kt
anotheroneofthese: We can make it a variable, but it does not seem like we can make it a property (as properties are instantiated/injected on the class, but aren't static).

MormonJesus69420 (approved):

src/test/kotlin/no/nb/mlt/wls/TestcontainerInitializer.kt
anotheroneofthese: I don't think there is too much we can do here for now, as we will require the container for future PR's

README.md
```diff
@@ -95,6 +95,8 @@ docker buildx build --platform linux/amd64 -t wls:latest docker/
 ***Caveats:***
 - When building the Docker image outside of NLNs network the build will fail, as it won't be able to access internal Harbor instance.
   In this case change the `FROM` line in the [Dockerfile](docker/Dockerfile "Link to project's Dockerfile") to `FROM eclipse-temurin:21-jdk-alpine` and build the image locally.
+- If you are outside of NLNs network your tests will fail, as an image for the dummy SynQ server is required for tests.
+  There is currently no replacement available for this.
```
MormonJesus69420, line 99:
```suggestion
  There is currently no replacement available for this, except for building the image for Dummy SynQ manually and pointing to it in places where it's used.
```

README.md
```diff
@@ -224,6 +226,7 @@ Although it's better to run them on your machine before pushing as it will save
 The following environment variables are relevant to configuring the application:
 
 - `KEYCLOAK_ISSUER_URI`: Is used to point at the Keycloak server used for authentication (required)
+- `KEYCLOAK_TOKEN_AUD`: Is used to set the audience for which Keycloak server the application intends to authenticate from (required)
```
MormonJesus69420, line 229:
```suggestion
- `KEYCLOAK_TOKEN_AUD`: Is used to set the audience of the Keycloak JWT token, it must match with the issued token audience value, which is different between environments (required)
```

MormonJesus69420 (changes requested):

MormonJesus69420 (approved):

PR #27, opened 2024-08-16
MLT-0040 More reactive handling

anotheroneofthese:
This PR refactors the codebase to achieve even further reactive handling. It makes the following improvements:
- Properly throw exceptions when the service times out
- Unify the handling of errors from SynQ between all services
- Streamline exception handling using more reactive code

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -113,6 +113,12 @@
     </dependency>
 
     <dependency>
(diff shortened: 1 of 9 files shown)


MormonJesus69420 (approved):
Looks good

PR #28, opened 2024-08-19
MLT-0032 Tests for OrderService

anotheroneofthese:
Adds a some tests for OrderService

Has one breaking change with renaming `getByHostNameAndHostOrderId` to `findByHostNameAndHostOrderId` in OrderRepository, to be more consistent with ProductRepository

### DIFF
--- a/src/main/kotlin/no/nb/mlt/wls/order/service/OrderService.kt
+++ b/src/main/kotlin/no/nb/mlt/wls/order/service/OrderService.kt
@@ -66,7 +66,7 @@ class OrderService(val db: OrderRepository, val synqService: SynqOrderService) {
             throw ServerWebInputException("The order's hostOrderId is required, and can not be blank")
         }
         if (payload.productLine.isEmpty()) {
-            throw ServerWebInputException("The order does not contain any product lines, and is therefore invalid")
+            throw ServerWebInputException("The order must contain product lines")
         }
     }
 }
--- a/src/main/kotlin/no/nb/mlt/wls/order/service/OrderService.kt
+++ b/src/main/kotlin/no/nb/mlt/wls/order/service/OrderService.kt
@@ -26,9 +26,13 @@ class OrderService(val db: OrderRepository, val synqService: SynqOrderService) {
         val existingOrder =
             db.findByHostNameAndHostOrderId(payload.hostName, payload.orderId)
                 .timeout(Duration.ofSeconds(8))
-                .onErrorMap(TimeoutException::class.java) {
-                    logger.error(it) {
-                        "Timed out while fetching from WLS database. Relevant payload: $payload"
+                .onErrorMap {
+                    if (it is TimeoutException) {
+                        logger.error(it) {
+                            "Timed out while fetching from WLS database. Relevant payload: $payload"
+                        }
+                    } else {
+                        logger.error(it) { "Unexpected error for $payload" }
                     }
                     ServerErrorException("Failed while checking if order already exists in the database", it)
                 }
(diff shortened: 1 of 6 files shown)


src/main/kotlin/no/nb/mlt/wls/order/service/OrderService.kt
```diff
@@ -35,7 +38,12 @@ class OrderService(val db: OrderRepository, val synqService: SynqOrderService) {
             return ResponseEntity.badRequest().build()
         }
 
-        synqService.createOrder(payload.toOrder().toSynqPayload())
+        // Handle any edge-cases
+        val synqResponse = synqService.createOrder(payload.toOrder().toSynqPayload())
+        if (!synqResponse.statusCode.isSameCodeAs(HttpStatus.CREATED)) {
+            throw ServerErrorException("Unexpected error with SynQ", null)
```
MormonJesus69420, line 44: Need to talk about it with Tom, how should we handle it

src/test/kotlin/no/nb/mlt/wls/order/service/OrderServiceTest.kt
```diff
@@ -0,0 +1,143 @@
...
+        assertExceptionThrownWithMessage(top.copy(orderId = "\t\n"), "orderId is required", ServerWebInputException::class.java)
+        assertExceptionThrownWithMessage(top.copy(orderId = "      "), "orderId is required", ServerWebInputException::class.java)
+    }
+
+    @Test
+    fun `save called with payload missing hostOrderId throws`() {
+        assertExceptionThrownWithMessage(top.copy(hostOrderId = ""), "The order's hostOrderId is required", ServerWebInputException::class.java)
+        assertExceptionThrownWithMessage(top.copy(hostOrderId = "\t\n"), "The order's hostOrderId is required", ServerWebInputException::class.java)
+        assertExceptionThrownWithMessage(top.copy(hostOrderId = "      "), "The order's hostOrderId is required", ServerWebInputException::class.java)
+    }
```
MormonJesus69420, line 56: What about orders with no productLines?

src/main/kotlin/no/nb/mlt/wls/order/service/OrderService.kt
anotheroneofthese: To clarify this edge-case only happens if the `createOrder` call doesn't throw, but still returns something that isn't a new order. Will update the comment accordingly

src/test/kotlin/no/nb/mlt/wls/order/service/OrderServiceTest.kt
```diff
@@ -0,0 +1,143 @@
...
+        runTest {
+            every { db.findByHostNameAndHostOrderId(top.hostName, top.hostOrderId) } returns Mono.empty()
+            coEvery { synq.createOrder(any()) } throws ServerErrorException("Duplicate order found in in SynQ", null)
+
+            assertExceptionThrownWithMessage(top, "Duplicate order", ServerErrorException::class.java)
+        }
+    }
+
+    @Test
+    fun `save called when synq fails is handled gracefully`() {
```
MormonJesus69420, line 77: It's mister SynQ for you, son

MormonJesus69420 (changes requested):

src/test/kotlin/no/nb/mlt/wls/order/service/OrderServiceTest.kt
anotheroneofthese: This now has a test

src/main/kotlin/no/nb/mlt/wls/order/service/OrderService.kt
MormonJesus69420: Yeah, just thinking about what to give back to caller and how to handle it in DB and storage systems, eg cleanup

PR #29, opened 2024-08-20
MLT-41 Change handling of duplicate orders

MormonJesus69420:
This pull requests aims to make handling of duplicate orders similar to the way we handle duplicate products.
It comes with the same weakness in case SynQ informs us that the order is a duplicate, not our DB. Problem being we have no information about the order and no way of getting that information through API.

### DIFF
--- a/src/main/kotlin/no/nb/mlt/wls/product/service/ProductService.kt
+++ b/src/main/kotlin/no/nb/mlt/wls/product/service/ProductService.kt
@@ -36,6 +36,7 @@ class ProductService(val db: ProductRepository, val synqProductService: SynqProd
                     ServerErrorException("Failed to save product in the database", it)
                 }
                 .awaitSingleOrNull()
+
         if (existingProduct != null) {
             return ResponseEntity.ok(existingProduct.toApiPayload())
         }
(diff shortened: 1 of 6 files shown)


MormonJesus69420:
I forgor about the API Doc...

src/main/kotlin/no/nb/mlt/wls/product/service/ProductService.kt
```diff
@@ -47,7 +48,13 @@ class ProductService(val db: ProductRepository, val synqProductService: SynqProd
         val product = payload.toProduct().copy(quantity = 0.0, location = null)
 
         // Product service should create the product in the storage system, and return error message if it fails
-        if (synqProductService.createProduct(product.toSynqPayload()).statusCode.isSameCodeAs(HttpStatus.OK)) {
+        val synqResponse = synqProductService.createProduct(product.toSynqPayload())
+        // If SynQ didn't throw an error, but returned 4xx/5xx, then it is likely some error or edge-case we haven't handled
+        if (synqResponse.statusCode.is4xxClientError || synqResponse.statusCode.is5xxServerError) {
+            throw ServerErrorException("Unexpected error with SynQ", null)
+        }
```
anotheroneofthese, line 55: This code might be redundant, as the WebClient should always map a 4xx or 5xx to an error, which is thrown

anotheroneofthese (commented):
Looks good. Only a single comment

src/main/kotlin/no/nb/mlt/wls/product/service/ProductService.kt
MormonJesus69420: If that's so, we might use it to handle 3xx codes and/or codes other than 200 and 201?

anotheroneofthese: I think codes other than what we expect (200 and 201) should be good

anotheroneofthese (approved):

PR #30, opened 2024-08-22
MLT-0042 Order GET endpoint

anotheroneofthese:
Adds the functionality to get orders from our database.
The assumptions are that you are authenticated with the correct hostname, and that you have an existing order ID

### DIFF
--- a/src/main/kotlin/no/nb/mlt/wls/order/controller/OrderController.kt
+++ b/src/main/kotlin/no/nb/mlt/wls/order/controller/OrderController.kt
@@ -77,6 +77,37 @@ class OrderController(val orderService: OrderService) {
         @RequestBody payload: ApiOrderPayload
     ): ResponseEntity<ApiOrderPayload> = orderService.createOrder(payload)
 
+    @Operation(
+        summary = "Gets an order from the storage system",
+        description = "Checks if a specified order exists within Hermes WLS."
+    )
+    @ApiResponses(
+        ApiResponse(
+            responseCode = "200",
+            description = "Order with given 'hostName' and 'hostOrderId' exists in the system.",
+            content = [
+                Content(
+                    mediaType = "application/json",
+                    schema = Schema(implementation = ApiOrderPayload::class)
+                )
+            ]
+        ),
+        ApiResponse(
+            responseCode = "400",
+            description = """Some field was invalid in your request. The error message contains information about the invalid fields.""",
+            content = [Content(schema = Schema())]
+        ),
+        ApiResponse(
+            responseCode = "401",
+            description = "Client sending the request is not authorized to request orders.",
+            content = [Content(schema = Schema())]
+        ),
+        ApiResponse(
+            responseCode = "403",
+            description = "A valid 'Authorization' header is missing from the request.",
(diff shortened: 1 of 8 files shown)


src/main/kotlin/no/nb/mlt/wls/order/service/OrderService.kt
```diff
@@ -76,4 +80,20 @@ class OrderService(val db: OrderRepository, val synqService: SynqOrderService) {
...
+
+    suspend fun getOrder(
+        hostName: HostName,
+        hostOrderId: String
+    ): ResponseEntity<Order> {
+        val order =
+            db.findByHostNameAndHostOrderId(hostName, hostOrderId)
+                .onErrorMap(IncorrectResultSizeDataAccessException::class.java) {
+                    TODO("Handle duplicate order ID's somehow")
+                }
```
MormonJesus69420, line 92: With change to DB so that hostName and hostOrderId are in a unique index, we avoid the issue of duplicate orders
That will of course require us to make sure that we don't duplicate orders when we create them

src/main/kotlin/no/nb/mlt/wls/order/controller/OrderController.kt
```diff
@@ -72,4 +76,41 @@ class OrderController(val orderService: OrderService) {
     suspend fun createOrder(
         @RequestBody payload: ApiOrderPayload
     ): ResponseEntity<ApiOrderPayload> = orderService.createOrder(payload)
+
+    @Operation(
+        summary = "Gets an order from the storage system",
+        description = "Checks if a specified order exists within Hermes WLS."
```
MormonJesus69420, line 82: Should probably expand on the API doc, but since I like doing it, I can fix it later on

src/main/kotlin/no/nb/mlt/wls/order/controller/OrderController.kt
```diff
@@ -72,4 +76,41 @@ class OrderController(val orderService: OrderService) {
...
+            responseCode = "403",
+            description = "A valid 'Authorization' header is missing from the request.",
+            content = [Content(schema = Schema())]
+        )
+    )
+    @GetMapping("/order/{hostName}/{hostOrderId}")
+    suspend fun getOrder(
+        @PathVariable("hostName") hostName: HostName,
+        @PathVariable("hostOrderId") hostOrderId: String
+    ): ResponseEntity<Order> = orderService.getOrder(hostName, hostOrderId)
```
MormonJesus69420, line 115: Should change it so that we get a JWT token and can use it to only show user orders that he owns

MormonJesus69420 (changes requested):
Looks good. Only the second comment, about the order ownership check requires changes

src/main/kotlin/no/nb/mlt/wls/order/service/OrderService.kt
anotheroneofthese: This error handling has been removed, as we still get the error if it ever occurs

MormonJesus69420 (approved):
Looks good, some small stuff here and there, but it's gonna be easier to fix and streamline it when we have all endpoints in order

PR #31, opened 2024-08-26
MLT-0043 Update Order (PUT) endpoint

anotheroneofthese:
Adds an API endpoint for updating existing orders

### DIFF
--- a/src/main/kotlin/no/nb/mlt/wls/order/controller/OrderController.kt
+++ b/src/main/kotlin/no/nb/mlt/wls/order/controller/OrderController.kt
@@ -160,6 +160,7 @@ class OrderController(val orderService: OrderService) {
     )
     @PutMapping("/order")
     suspend fun updateOrder(
+        @AuthenticationPrincipal jwt: JwtAuthenticationToken,
         @RequestBody payload: ApiOrderPayload
-    ): ResponseEntity<ApiOrderPayload> = orderService.updateOrder(payload)
+    ): ResponseEntity<ApiOrderPayload> = orderService.updateOrder(payload, jwt.name)
 }
--- a/src/main/kotlin/no/nb/mlt/wls/order/controller/OrderController.kt
+++ b/src/main/kotlin/no/nb/mlt/wls/order/controller/OrderController.kt
@@ -53,7 +53,7 @@ class OrderController(val orderService: OrderService) {
         ),
         ApiResponse(
             responseCode = "400",
-            description = """Order payload is invalid and was not updated.
+            description = """Order payload is invalid and was not created.
                 An empty error message means the order already exists with the current ID.
                 Otherwise, the error message contains information about the invalid fields.""",
             content = [Content(schema = Schema())]
--- a/src/main/kotlin/no/nb/mlt/wls/order/controller/OrderController.kt
+++ b/src/main/kotlin/no/nb/mlt/wls/order/controller/OrderController.kt
@@ -72,4 +72,8 @@ class OrderController(val orderService: OrderService) {
     suspend fun createOrder(
         @RequestBody payload: ApiOrderPayload
     ): ResponseEntity<ApiOrderPayload> = orderService.createOrder(payload)
+
+    suspend fun updateOrder(
+        @RequestBody payload: ApiOrderPayload
+    ): ResponseEntity<ApiOrderPayload> = orderService.updateOrder(payload)
 }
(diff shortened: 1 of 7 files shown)


src/main/kotlin/no/nb/mlt/wls/order/controller/OrderController.kt
```diff
@@ -72,4 +73,48 @@ class OrderController(val orderService: OrderService) {
...
+        ApiResponse(
+            responseCode = "409",
+            description = "The order is already being processed, and can not be edited at this point.",
+            content = [Content(schema = Schema())]
+        )
+    )
+    @PutMapping("/order")
+    suspend fun updateOrder(
+        @RequestBody payload: ApiOrderPayload
+    ): ResponseEntity<ApiOrderPayload> = orderService.updateOrder(payload)
```
MormonJesus69420, line 164: Check if user is permitted to update order, same as others

src/main/kotlin/no/nb/mlt/wls/order/model/Order.kt
```diff
@@ -4,12 +4,18 @@ import io.swagger.v3.oas.annotations.media.Schema
 import io.swagger.v3.oas.annotations.media.Schema.AccessMode.READ_ONLY
 import no.nb.mlt.wls.core.data.HostName
 import no.nb.mlt.wls.core.data.Owner
+import org.bson.types.ObjectId
+import org.springframework.data.annotation.Id
+import org.springframework.data.mongodb.core.index.CompoundIndex
 import org.springframework.data.mongodb.core.mapping.Document
 
+@CompoundIndex(unique = true, def = "{'hostName':1,'hostOrderId':1}")
```
MormonJesus69420, line 12: Should also have it for products, I think. Eg. hostName and productId, since they also should be unique

src/main/kotlin/no/nb/mlt/wls/order/service/OrderService.kt
```diff
@@ -65,6 +67,53 @@ class OrderService(val db: OrderRepository, val synqService: SynqOrderService) {
...
+                    if (it is TimeoutException) {
+                        logger.error(it) {
+                            "Timed out while fetching from WLS database. Relevant payload: $payload"
+                        }
+                    } else {
+                        logger.error(it) { "Unexpected error for $payload" }
+                    }
+                    ServerErrorException("Failed while checking if order already exists in the database", it)
+                }
+                .awaitSingleOrNull()
```
MormonJesus69420, line 96: I think it's better to extract that into a method. I have done it in my branch, just need to remember to replace these calls when we merge them.

src/main/kotlin/no/nb/mlt/wls/order/service/OrderService.kt
```diff
@@ -65,6 +67,53 @@ class OrderService(val db: OrderRepository, val synqService: SynqOrderService) {
...
+        val updatedOrder =
+            db.save(
+                existingOrder.copy(
+                    hostOrderId = payload.hostOrderId,
+                    hostName = payload.hostName,
+                    productLine = payload.productLine,
+                    orderType = payload.orderType,
+                    owner = payload.owner,
+                    receiver = payload.receiver,
+                    callbackUrl = payload.callbackUrl
```
MormonJesus69420, line 120: I think I mentioned how update order should have its own payload which only takes in fields that can be updated, it doesn't make sense to change hostOrderId, hostName, owner, or callbackUrl. The order update payload should only take in productline, orderType, and receiver, allowing users to only change these values.

MormonJesus69420 (changes requested):

src/main/kotlin/no/nb/mlt/wls/order/service/OrderService.kt
anotheroneofthese: Same

anotheroneofthese:
Marking this as ready for now. We do still need Controller tests, but to do this we first need to update dummy-synq

src/main/kotlin/no/nb/mlt/wls/order/model/Order.kt
anotheroneofthese: This change should be easy to implement in the DB and the code. I can include it in this PR if desirable

src/main/kotlin/no/nb/mlt/wls/order/service/OrderService.kt
```diff
@@ -76,6 +64,41 @@ class OrderService(val db: OrderRepository, val synqService: SynqOrderService) {
...
+            throw ResponseStatusException(HttpStatus.CONFLICT, "Order is already being processed, and can not be edited")
+        }
+        synqService.updateOrder(payload)
+
+        // Saving here will override the existing order, as the id's match
+        val updatedOrder =
+            db.save(
+                existingOrder.copy(
+                    hostOrderId = payload.hostOrderId,
+                    hostName = payload.hostName,
```
MormonJesus69420, line 91:
```suggestion
```

MormonJesus69420 (approved):

PR #33, opened 2024-09-05
Controller Test Improvements

anotheroneofthese:
A small thing on the side I have been preparing until all the endpoints were merged into main. This contains tests for the GET, UPDATE, and DELETE endpoints.

### DIFF
--- a/src/test/kotlin/no/nb/mlt/wls/order/controller/OrderControllerTest.kt
+++ b/src/test/kotlin/no/nb/mlt/wls/order/controller/OrderControllerTest.kt
@@ -59,7 +59,7 @@ class OrderControllerTest(
 
     private lateinit var webTestClient: WebTestClient
 
-    val clientName: String = HostName.AXIELL.name
+    val client: String = HostName.AXIELL.name
 
     @BeforeEach
     fun setUp() {
@@ -83,7 +83,7 @@ class OrderControllerTest(
 
             webTestClient
                 .mutateWith(csrf())
-                .mutateWith(mockJwt().jwt { it.subject(clientName) })
+                .mutateWith(mockJwt().jwt { it.subject(client) })
                 .post()
                 .accept(MediaType.APPLICATION_JSON)
                 .bodyValue(testOrderPayload)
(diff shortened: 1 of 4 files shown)


src/test/kotlin/no/nb/mlt/wls/order/controller/OrderControllerTest.kt
```diff
@@ -164,6 +167,86 @@ class OrderControllerTest(
             .expectStatus().is5xxServerError
     }
 
+    // TODO - Should this endpoint really be returning Orders directly?
```
MormonJesus69420, line 170: Nope, DTOs

src/test/kotlin/no/nb/mlt/wls/order/controller/OrderControllerTest.kt
```diff
@@ -175,16 +258,31 @@ class OrderControllerTest(
                 .mutateWith(csrf())
                 .mutateWith(mockJwt().jwt { it.subject("axiell") })
                 .delete()
-                .uri("/${duplicateOrderPayload.hostName}/${duplicateOrderPayload.hostOrderId}")
+                .uri("/${dop.hostName}/${dop.hostOrderId}")
```
MormonJesus69420, line 261: Dunno which method of making URI is better/prefrerred/standard/etc, this or the one on line 177. If it's either or, then we should pick one and stick to it.

src/test/kotlin/no/nb/mlt/wls/order/controller/OrderControllerTest.kt
```diff
@@ -211,8 +309,9 @@ class OrderControllerTest(
             callbackUrl = "callbackUrl"
         )
 
+    // dop = Duplicate Order Payload
     // Will exist in the database
-    private val duplicateOrderPayload =
+    private val dop =
```
MormonJesus69420, line 314: If you've changed it to `dop` then we might as well rename `testOrderPayload` to `top` and probably do it in all the test files we have. Also best to add the `// dop = Duplicate Order Payload` as a JavaDoc so that Intellij shows it in case someone forgets what the abbreviation stands for.
I personally prefer to only use abbreviations when names are extra long or the abbreviation is well known, such as `cut = Class Under Test`, as it might make code harder to read and reason about otherwise. I like it when code can be easily read in plain text editor/terminal/etc without any fancy help tools. You might also check with Tom Kristian to see what his opinion is.

MormonJesus69420 (changes requested):
Good job so far, and good on noticing the GET endpoint returning a pure Order object. Makes me fell silly for missing that in previous review

src/test/kotlin/no/nb/mlt/wls/order/controller/OrderControllerTest.kt
anotheroneofthese: I made that change one day as I believed it to be inconsistent. This was only true when looking at the service tests (OrderServiceTest and ProductServiceTest). If we do make this change I will ensure everything has relevant JavaDoc, but I do agree that when reading the code for review it would be better if all things were abbreviated

src/test/kotlin/no/nb/mlt/wls/order/controller/OrderControllerTest.kt
anotheroneofthese: Will make a separate PR to resolve this

src/test/kotlin/no/nb/mlt/wls/order/controller/OrderControllerTest.kt
anotheroneofthese: Using var args to make it more consistent across the codebase

src/test/kotlin/no/nb/mlt/wls/order/controller/OrderControllerTest.kt
anotheroneofthese: After chatting with Tom we agreed on making the code more readable by expanding the abbrevations

tomolse (approved):

PR #36, opened 2024-09-27
Fix not being able to send orders to SynQ

anotheroneofthese:
Some of the fields in our data structures are not required for our use-cases, and prevents us from creating new orders

### DIFF
--- a/src/main/kotlin/no/nb/mlt/wls/infrastructure/synq/SynqOrderPayload.kt
+++ b/src/main/kotlin/no/nb/mlt/wls/infrastructure/synq/SynqOrderPayload.kt
@@ -21,8 +21,7 @@ data class SynqOrderPayload(
     val priority: Int,
     val owner: SynqOwner,
     val orderLine: List<OrderLine>,
-    val customer: String? = null,
-    val shippingAddress: ShippingAddress? = null
+    val customer: String? = null
 ) {
     data class OrderLine(
         @Min(1)
@@ -30,21 +29,6 @@ data class SynqOrderPayload(
         val productId: String,
         val quantityOrdered: Double
     )
-
-    data class ShippingAddress(
-        val address: Address?
-    ) {
-        data class Address(
-            val contactPerson: String? = null,
-            val addressLine1: String? = null,
-            val city: String? = null,
-            val state: String? = null,
-            val postalCode: String? = null,
-            val country: String? = null,
-            val phoneNumber: String? = null,
-            val email: String? = null
-        )
-    }
 
     enum class SynqOrderType(private val type: String) {
         STANDARD("Standard");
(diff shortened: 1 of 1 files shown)


src/main/kotlin/no/nb/mlt/wls/infrastructure/synq/SynqOrderPayload.kt
```diff
@@ -21,8 +21,7 @@ data class SynqOrderPayload(
     val priority: Int,
     val owner: SynqOwner,
     val orderLine: List<OrderLine>,
-    val customer: String? = null,
-    val shippingAddress: ShippingAddress? = null
+    val customer: String? = null
```
MormonJesus69420, line 24: As of now customer field does not seem to work in SynQ, additionally it requires the customer to be created beforehand and we don't seem to have support for creating it with API.
Therefore, I vote to have it removed from payload. Additionally, we should include address, even though it is only name and street address. SynQ seems to accept a shipping address with only those two fields (`contactPerson` and `addressLine1`):
image

src/main/kotlin/no/nb/mlt/wls/infrastructure/synq/SynqOrderPayload.kt
```diff
@@ -75,19 +59,7 @@ fun Order.toSynqPayload() =
...
-fun Order.Receiver.toShippingAddress(): SynqOrderPayload.ShippingAddress =
-    SynqOrderPayload.ShippingAddress(
-        SynqOrderPayload.ShippingAddress.Address(
-            contactPerson = name,
-            addressLine1 = address,
-            city = city,
-            postalCode = postalCode,
-            phoneNumber = phoneNumber
-        )
+        customer = receiver.name
```
MormonJesus69420, line 62: As mentioned above, drop customer, add back shippingAddress

src/test/kotlin/no/nb/mlt/wls/domain/WLSServiceTest.kt
```diff
@@ -236,7 +236,7 @@ class WLSServiceTest {
                     "12345",
                     listOf("mlt-420", "mlt-421"),
                     Order.Type.LOAN,
-                    Order.Receiver("name", "location", null, null, null, null),
+                    Order.Receiver("name", "location"),
```
MormonJesus69420, line 239: ... and don't forget to update the tests :)

MormonJesus69420 (changes requested):

src/test/kotlin/no/nb/mlt/wls/order/model/MongoOrderModelConversionTest.kt
```diff
@@ -46,11 +43,7 @@ class MongoOrderModelConversionTest {
             receiver =
                 Order.Receiver(
                     name = "name",
-                    address = "address",
-                    postalCode = "postalCode",
-                    city = "city",
-                    phoneNumber = "phoneNumber",
-                    location = "location"
+                    address = "location"
```
MormonJesus69420, line 46:
```suggestion
                    address = "address"
```

MormonJesus69420 (approved):
Should be good now and comply with the mock API after changes

PR #37, opened 2024-10-03
MLT-0053 Add use-case for moving items

anotheroneofthese:
Originally planned as updating items - this PR implements the functionality to move items within WLS. This specific use-case is required for when the storage systems gives us various status updates when the items are withdrawn, inserted, or generally moved around

### DIFF
--- a/src/main/kotlin/no/nb/mlt/wls/domain/ports/inbound/MoveItem.kt
+++ b/src/main/kotlin/no/nb/mlt/wls/domain/ports/inbound/MoveItem.kt
@@ -1,0 +1,21 @@
+package no.nb.mlt.wls.domain.ports.inbound
+
+import no.nb.mlt.wls.domain.model.HostName
+import no.nb.mlt.wls.domain.model.Item
+
+/**
+ * This port is used for handling messages regarding items moving
+ * inside the storage system.
+ * Some examples include handling status messages when crates arrive
+ * at picking stations, or when items return to the
+ * storage systems.
+ * In both cases we want to know where the item went, and if the count changed
+ */
+interface MoveItem {
+    suspend fun moveItem(
+        hostId: String,
+        hostName: HostName,
+        quantity: Double,
+        location: String
+    ): Item
+}
--- a/src/main/kotlin/no/nb/mlt/wls/domain/ports/outbound/ItemRepository.kt
+++ b/src/main/kotlin/no/nb/mlt/wls/domain/ports/outbound/ItemRepository.kt
@@ -12,7 +12,7 @@ interface ItemRepository {
 
     fun createItem(item: Item): Mono<Item>
 
-    suspend fun doesAllItemsExist(ids: List<ItemId>): Boolean
+    suspend fun doesEveryItemExist(ids: List<ItemId>): Boolean
 
     suspend fun moveItem(
         hostId: String,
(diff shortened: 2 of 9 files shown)


src/main/kotlin/no/nb/mlt/wls/domain/ports/outbound/ItemRepository.kt
```diff
@@ -13,6 +13,15 @@ interface ItemRepository {
     fun createItem(item: Item): Mono<Item>
 
     suspend fun doesAllItemsExist(ids: List<ItemId>): Boolean
+
```
MormonJesus69420, line 16: I know it's unrelated, but can we fix the grammar of the above metod to "doAllItemsExist" or "doesEveryItemExist"?

src/main/kotlin/no/nb/mlt/wls/domain/ports/inbound/MoveItem.kt
```diff
@@ -0,0 +1,21 @@
...
+ * Some examples include handling status messages when crates arrive
+ * at picking stations, or when items return to the
+ * storage systems.
+ * In both cases we want to know where the item went, and if the count changed
+ */
+interface MoveItem {
+    suspend fun moveItem(
+        hostId: String,
+        hostName: HostName,
+        quantity: Double,
```
MormonJesus69420, line 18: Don't know where the best place is to check, but quantity should be zero or positive, would be funky if we have -1 book.

src/main/kotlin/no/nb/mlt/wls/domain/ports/inbound/MoveItem.kt
```diff
@@ -0,0 +1,21 @@
...
+        location: String
```
MormonJesus69420, line 19: Should check if location is not a blank value. Could probably do that and the above check with annotations, or do a logic check somewhere else in the logic flow, probably service if I were to make the call.

Which reminds me that we should also make sure that when an item is registered the location is set to a null, but only then, cause it's the only case where location can be null, at least so far.

MormonJesus69420 (changes requested):
Looks good for now, just a few minor niggles to iron out.

Also, we need to keep in mind that in future we will need to do some more work when item moves.

For example, checking if item moves into the storage system, and is part of an order, so we can update the item as returned. Will have to discuss how to best model and implement that kind of logic.

Other than that, it's looking great!

src/main/kotlin/no/nb/mlt/wls/domain/ports/inbound/MoveItem.kt
MormonJesus69420: With those changes made, we should also have tests to validate that of course.

anotheroneofthese: For the latter: Items are registered using `ItemMetadata`, which when mapped to `Item` sets these fields to null. They will only be mapped if you explictly set them. If you'd like I could try writing a comment which describes this intention

src/test/kotlin/no/nb/mlt/wls/domain/WLSServiceTest.kt
```diff
@@ -114,8 +115,77 @@ class WLSServiceTest {
...
+    }
+
+    @Test
+    fun `moveItem throws when count is invalid`() {
+        coEvery { itemRepoMock.moveItem(any(), any(), -1.0, any()) } throws ValidationException("Location cannot be blank")
+
+        val cut = WLSService(itemRepoMock, orderRepoMock, storageSystemRepoMock)
+        runTest {
+            assertThrows<RuntimeException> {
+                cut.moveItem(testItem.hostId, testItem.hostName, -1.0, "   ")
```
MormonJesus69420, line 167:
```suggestion
                cut.moveItem(testItem.hostId, testItem.hostName, -1.0, "Somewhere nice")
```

MormonJesus69420 (approved):
Looks great!

PR #39, opened 2024-10-04
MLT-0038 Callback for Orders

MormonJesus69420:
Implements a system to handle status updates from SynQ and make callbacks to Axiell. Done of course in a way that should be extensible so we can easily add more storages and catalogues.


MormonJesus69420:
In regards to the architecture test failure: not sure how to handle that. Do we move the callback handling to application layer, do we split responsiblity for it somehow, or do I just need to create a callback item/order models and send those instead?
I think the last option makes most sense, as technically we might in the future send different data on callback than what we get into the host API? What do the others think about it?

anotheroneofthese:
> In regards to the architecture test failure: not sure how to handle that. Do we move the callback handling to application layer, do we split responsiblity for it somehow, or do I just need to create a callback item/order models and send those instead? I think the last option makes most sense, as technically we might in the future send different data on callback than what we get into the host API? What do the others think about it?

I also like the latter option more (implementing callback models), albeit I feel at the moment we don't have any clear use-cases for them in the future. Arguably we could move the callbacks into `application`, as they are an internal part of the app who will send messages out via the ports. Regardless, we should await a response from @tomolse

MormonJesus69420:
Thinking of merging this now. I know the tests are not done, but this is pretty much a blocking PR and I want to give @anotheroneofthese code to work with.

Suggest we just merge this as is, and I open new PR to add testing to the new features and then add another one for the other order callback.

MormonJesus69420:
image
Spotless go home you are drunk

anotheroneofthese (approved):

tomolse (approved):

PR #43, opened 2024-10-14
MLT-0055 Item Picking

anotheroneofthese:
This implements functionality where items are able to be picked out of WLS from various storage systems. For now our use-cases cover SynQ


src/main/kotlin/no/nb/mlt/wls/application/synqapi/synq/SynqController.kt
```diff
@@ -52,13 +58,43 @@ class SynqController(
         )
     )
     @PutMapping("/item-update")
-    suspend fun updateItem(
+    suspend fun moveItem(
```
MormonJesus69420, line 61: Maybe rename the endpoint to /move-item as well? That way it will be more consistent with function name and define the scope better.

src/main/kotlin/no/nb/mlt/wls/application/synqapi/synq/SynqController.kt
```diff
@@ -52,13 +58,43 @@ class SynqController(
...
+    )
+    @PutMapping("/pick-update/{owner}/{orderId}")
+    suspend fun pickOrder(
+        @Parameter(description = "Owner of the order items")
+        @PathVariable owner: Owner,
+        @Parameter(description = "Order ID in the storage system")
+        @PathVariable orderId: String,
+        @RequestBody payload: SynqOrderPickingConfirmationPayload
+    ) {
+        if (payload.orderLine.isEmpty()) throw ValidationException("Picking update does not contain any elements in the order line")
```
MormonJesus69420, line 104: I started moving this kind of validation to the payloads themselves. I think it's a bit more logical as it's payloads job to know if it's valid, not controller or other. If you look at what I have done in the recent commit to my branch you can see it for ApiOrderPayload and ApiItemPayload.

src/main/kotlin/no/nb/mlt/wls/domain/WLSService.kt
```diff
@@ -76,6 +78,59 @@ class WLSService(
...
+            if (itemsPickedMap[item.hostId] == null) {
+                // TODO - Review this. This state should generally not happen with the preconditions. Does it need to be handled at all?
+                throw ItemNotFoundException("Item with ID ${itemId.hostId} for host ${item.hostName} was not found, despite existing!")
+            }
+            val movedItem =
+                itemRepository.moveItem(
+                    item.hostId,
+                    item.hostName,
+                    item.quantity?.minus((itemsPickedMap[item.hostId] ?: 0.0)) ?: 0.0,
+                    "Picked"
```
MormonJesus69420, line 105: Location should be WITH_LENDER or similar and only if quantity is 0.0, also shouldn't you check if quantity is negative after subtraction?

src/main/kotlin/no/nb/mlt/wls/domain/WLSService.kt
```diff
@@ -76,6 +78,59 @@ class WLSService(
...
+        hostIds: List<String>,
+        orderId: String
+    ) {
+        // Make a new order line with picked items
+        val order = getOrder(hostName, orderId) ?: throw OrderNotFoundException("Order $orderId for host $hostName not found")
+
+        val orderLine =
+            order.orderLine.map { orderItem ->
+                hostIds.map { hostId ->
+                    if (orderItem.hostId == hostId) {
```
MormonJesus69420, line 123: Wouldn't something along these lines work too?

```kotlin
val newOrderLines = order.orderLine.map {
    if (hostIds.contains(it.hostId) {
        it.copy(status = Order.OrderItem.Status.PICKED)
    } else {
        it
    }
}
```
That way you could keep the status as a val not var and I think it's a bit clearer what's going on.

src/main/kotlin/no/nb/mlt/wls/application/synqapi/synq/SynqController.kt
anotheroneofthese: I will take a look at that and see if i can't move this validation then

src/main/kotlin/no/nb/mlt/wls/domain/WLSService.kt
anotheroneofthese: Should we declare some statics for locations somewhere?
Also yes, we do need a check here. I forgot that this isn't being passed to WLSServices `moveItem` method. 
What should we do in the scenario where someone has picked more items than we have registered? Be it invalid state, or another mismatch in inventory

anotheroneofthese: I am handling this by logging the error and clamping the quantity. We should still consider implementing further warnings in case this happens (send some notification somewhere?)

src/main/kotlin/no/nb/mlt/wls/application/synqapi/synq/SynqOrderPickingConfirmationPayload.kt
```diff
@@ -44,7 +45,21 @@ data class SynqOrderPickingConfirmationPayload(
         example = "Sikringsmagasin_2"
     )
     val warehouse: String
-)
+) {
+    fun validate() {
+        if (operator.trim().isBlank()) {
+            throw ValidationException("Operator name can not be blank")
+        }
+        if (warehouse.trim().isBlank()) {
```
MormonJesus69420, line 53:
```suggestion
        if (warehouse.isBlank()) {
```

src/main/kotlin/no/nb/mlt/wls/application/synqapi/synq/SynqOrderPickingConfirmationPayload.kt
```diff
@@ -112,4 +127,16 @@ data class OrderLine(
         example = "[{...}]"
     )
     val attributeValue: List<AttributeValue>
-)
+) {
+    fun validate() {
+        if (productId.trim().isBlank()) {
```
MormonJesus69420, line 132:
```suggestion
        if (productId.isBlank()) {
```

src/main/kotlin/no/nb/mlt/wls/application/synqapi/synq/SynqOrderPickingConfirmationPayload.kt
```diff
@@ -112,4 +127,16 @@ data class OrderLine(
...
+            throw ValidationException("Product ID can not be blank")
+        }
+        if (hostName.trim().isBlank()) {
```
MormonJesus69420, line 135:
```suggestion
        if (hostName.isBlank()) {
```

src/main/kotlin/no/nb/mlt/wls/infrastructure/repositories/item/MongoItemRepository.kt
```diff
@@ -87,6 +87,24 @@ class ItemRepositoryMongoAdapter(
 
         return getItem(hostName, hostId)!!
     }
+
+    override suspend fun pickItem(item: Item) {
+        mongoRepo
+            .findAndUpdateItemByHostNameAndHostId(item.hostName, item.hostId, 0.0, "Picked")
```
MormonJesus69420, line 93: Shouldn't we either calculate the quantity here, or get item object with correct quantity?
Also Picked is not a location, we either keep the location as is (if quantity > 0) or WITH_LENDER (if quantity = 0)
The same rules for quantity and location apply to our DB as to what we return in callback to hosts.

src/main/kotlin/no/nb/mlt/wls/domain/WLSService.kt
```diff
@@ -76,6 +78,67 @@ class WLSService(
...
+                    "WITH_LENDER"
+                )
+            inventoryNotifier.itemChanged(movedItem)
+        }
+        // TODO - Should we log this once completed?
+    }
+
+    override suspend fun pickOrderItems(
+        hostName: HostName,
+        hostIds: List<String>,
```
MormonJesus69420, line 125: Just as a suggestion for better clarity, maybe change the name to something like pickedItemsId? Makes it more clear in the map function below what is going on

src/main/kotlin/no/nb/mlt/wls/domain/WLSService.kt
```diff
@@ -76,6 +78,67 @@ class WLSService(
...
+        if (!itemRepository.doesEveryItemExist(itemIds)) {
+            throw ItemNotFoundException("Some items do not exist in the database, and were unable to be picked")
+        }
+
+        itemIds.map { itemId ->
+            val item = getItem(itemId.hostName, itemId.hostId)!!
+            if (itemsPickedMap[item.hostId] == null) {
+                // TODO - Review this. This state should generally not happen with the preconditions. Does it need to be handled at all?
+                throw ItemNotFoundException("Item with ID ${itemId.hostId} for host ${item.hostName} was not found, despite existing!")
+            }
```
MormonJesus69420, line 99: I think this check can be safely removed

src/main/kotlin/no/nb/mlt/wls/domain/WLSService.kt
```diff
@@ -76,6 +78,67 @@ class WLSService(
...
+
+            val itemsStocked = item.quantity ?: 0.0
+            val itemsPicked = itemsPickedMap[item.hostId] ?: 0.0
```
MormonJesus69420, line 102: itemsInStockQuantity, pickedItemsQuantity

src/main/kotlin/no/nb/mlt/wls/domain/WLSService.kt
```diff
@@ -76,6 +78,67 @@ class WLSService(
...
+                logger.error {
+                    "Tried to pick too many items for ${item.hostId}. WLS DB has $itemsStocked stocked, and storage system tried to pick $itemsPicked"
+                }
+            }
+            val movedItem =
+                itemRepository.moveItem(
+                    item.hostId,
+                    item.hostName,
+                    Math.clamp(itemsPicked.minus(itemsStocked), 0.0, Double.MAX_VALUE),
+                    "WITH_LENDER"
```
MormonJesus69420, line 116: WITH_LENDER only if new quantity = 0.0

src/main/kotlin/no/nb/mlt/wls/domain/WLSService.kt
```diff
@@ -76,6 +78,67 @@ class WLSService(
...
+            if (itemsPicked > itemsStocked) {
+                logger.error {
+                    "Tried to pick too many items for ${item.hostId}. WLS DB has $itemsStocked stocked, and storage system tried to pick $itemsPicked"
+                }
+            }
+            val movedItem =
+                itemRepository.moveItem(
+                    item.hostId,
+                    item.hostName,
+                    Math.clamp(itemsPicked.minus(itemsStocked), 0.0, Double.MAX_VALUE),
```
MormonJesus69420, line 115: Shouldn't that be other way around: itemsStocked.minus(itemsPicked)?
Since move item function sets the new quantity and location, so it should be 'new quantity' = 'number of items in stock' - 'number of picked items'

src/main/kotlin/no/nb/mlt/wls/infrastructure/repositories/item/MongoItemRepository.kt
anotheroneofthese: This function is actually unused at the moment. The functionality for picking was delegated to use `moveItem` calls instead

src/main/kotlin/no/nb/mlt/wls/domain/WLSService.kt
anotheroneofthese: Correct

src/main/kotlin/no/nb/mlt/wls/domain/WLSService.kt
anotheroneofthese: As discussed, as it is a bit better for our internal use case, we will rename this to `pickedHostIds`

src/main/kotlin/no/nb/mlt/wls/domain/WLSService.kt
```diff
@@ -93,27 +93,38 @@ class WLSService(
...
+            if (pickedItemsQuantity > itemsInStockQuantity) {
                 logger.error {
-                    "Tried to pick too many items for ${item.hostId}. WLS DB has $itemsStocked stocked, and storage system tried to pick $itemsPicked"
+                    "Tried to pick too many items for ${item.hostId}. " +
+                        "WLS DB has $itemsInStockQuantity stocked, and storage system tried to pick $pickedItemsQuantity"
                 }
             }
+            val quantity = Math.clamp(itemsInStockQuantity.minus(pickedItemsQuantity), 0.0, Double.MAX_VALUE)
+            val location: String =
+                if (quantity == 0.0) {
```
MormonJesus69420, line 110: Not fully sure if this will ever happen, but since floating point arithmetics can be a bit fucky we might end up with quantity being something like 0.0000000001 which won't equal 0.0 according to this check. I am unsure if either the `minus` or `clamp` functions handle that in any way to prevent this scenario.
If we want to be super sure we can add a simple extension function to Double class like:
```kotlin
infix fun Double.eq(other: Double) = abs(this - other) < 0.000001
```
And then use it here as `quantity eq 0.0`

No need to do it right now, but it might be something to think about

MormonJesus69420 (approved):
Looks good besides a small comment I had about double comparison

src/main/kotlin/no/nb/mlt/wls/domain/WLSService.kt
MormonJesus69420: We can also use BigDecimal which should give us accurate results

MormonJesus69420 (approved):
lgtm

PR #44, opened 2024-10-18
MLT-0060 Create Swagger documentation for Item and Order callbacks

anotheroneofthese:
Adds callback documentation to the controllers, explaining what the response looks like.
One remaining question is whether we add this to some of the other controllers. Currently the only endpoints that produce the callbacks have received the annotation. Should the `createOrder` and `createItem` also be annotated?


MormonJesus69420:
Looks good, however this documentation should be put on the createOrder and createItem endpoints instead, as it's actually those that have callbacks. You can see that in our mock API

MormonJesus69420 (approved):

PR #45, opened 2024-10-21
MLT-0061 Delete Orders from Storage Systems

anotheroneofthese:
This should also resolve MLT-0062.
The PR expands the port (use-case) of the delete endpoint to use the order model. Alternatively this could be implemented to use a DTO to be more precise, but this should be fine for now in order to correctly delete orders from the storage systems.

### DIFF
--- a/src/main/kotlin/no/nb/mlt/wls/domain/WLSService.kt
+++ b/src/main/kotlin/no/nb/mlt/wls/domain/WLSService.kt
@@ -140,7 +140,7 @@ class WLSService(
         orderId: String
     ) {
         // Make a new order line with picked items
-        val order = getOrder(hostName, orderId) ?: throw OrderNotFoundException("Order $orderId for host $hostName not found")
+        val order = getOrderOrThrow(hostName, orderId)
         val orderLine =
             order.orderLine.map { orderItem ->
                 if (pickedHostIds.contains(orderItem.hostId)) {
@@ -179,7 +179,8 @@ class WLSService(
         hostName: HostName,
         hostOrderId: String
     ) {
-        storageSystemFacade.deleteOrder(hostName, hostOrderId)
+        val order = getOrderOrThrow(hostName, hostOrderId)
+        storageSystemFacade.deleteOrder(order)
         orderRepository.deleteOrder(hostName, hostOrderId)
     }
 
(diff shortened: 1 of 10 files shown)


src/main/kotlin/no/nb/mlt/wls/domain/WLSService.kt
```diff
@@ -242,4 +245,15 @@ class WLSService(
...
+    @Throws(OrderNotFoundException::class)
+    suspend fun getOrderOrThrow(
+        hostName: HostName,
+        hostOrderId: String
+    ): Order {
+        return getOrder(
+            hostName,
+            hostOrderId
+        ) ?: throw OrderNotFoundException("No order with hostOrderId: $hostOrderId and hostName: $hostName exists")
+    }
```
tomolse, line 258: This could be a private function.

src/main/kotlin/no/nb/mlt/wls/domain/WLSService.kt
```diff
@@ -196,11 +201,9 @@ class WLSService(
...
         }
 
-        val order =
-            getOrder(
-                hostName,
-                hostOrderId
-            ) ?: throw OrderNotFoundException("No order with hostOrderId: $hostOrderId and hostName: $hostName exists")
+        val order = getOrderOrThrow(hostName, hostOrderId)
+
+        order.throwIfInProgress()
```
tomolse, line 206: This validation should be handled in the Order class (I think it already does)

src/main/kotlin/no/nb/mlt/wls/domain/WLSService.kt
```diff
@@ -139,8 +141,9 @@ class WLSService(
         pickedHostIds: List<String>,
         orderId: String
     ) {
+        // TODO - Move some of this to Order
         // Make a new order line with picked items
-        val order = getOrder(hostName, orderId) ?: throw OrderNotFoundException("Order $orderId for host $hostName not found")
+        val order = getOrderOrThrow(hostName, orderId)
```
tomolse, line 146: Validation should be done in the Order class.

tomolse (approved):
The comments in this review are going to be fixed in another PR.

PR #46, opened 2024-10-25
MLT-0065 Update Local DB test data

anotheroneofthese:
This is a small fix which fixes the test data being out of date. Also adds an extra item and a test order.


tomolse (approved):
LGTM

PR #48, opened 2024-10-29
MLT-0063 Use Ints for quantity

anotheroneofthese:
For our use-cases we do not need doubles for our Item models. This PR refactors this to use integers instead

### DIFF
--- a/src/main/kotlin/no/nb/mlt/wls/application/synqapi/synq/SynqBatchMoveItemPayload.kt
+++ b/src/main/kotlin/no/nb/mlt/wls/application/synqapi/synq/SynqBatchMoveItemPayload.kt
@@ -131,7 +131,7 @@ data class Product(
         example = "1.0"
     )
     @PositiveOrZero
-    val quantityOnHand: Double,
+    val quantityOnHand: Int,
     @Schema(
         description = "Signifies the product is marked as suspect in the storage system and needs to be manually verified.",
         example = "false"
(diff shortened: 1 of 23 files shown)


src/main/kotlin/no/nb/mlt/wls/application/synqapi/synq/SynqBatchMoveItemPayload.kt
```diff
@@ -127,11 +127,13 @@ data class Product(
     )
     val productVersionId: String,
     @Schema(
-        description = "Quantity of the product in the transport unit, uses float, however in our case we operate only in whole numbers.",
+        description =
+            "Quantity of the product in the transport unit. Accepts doubles, " +
+                "however, this is converted as we operate only in whole numbers.",
```
MormonJesus69420, line 132:
```suggestion
        description = """Quantity of the product in the transport unit. 
            Accepts doubles, but they will be converted to integers.""",
```

src/main/kotlin/no/nb/mlt/wls/domain/WLSService.kt
```diff
@@ -110,9 +110,9 @@ class WLSService(
                         "WLS DB has $itemsInStockQuantity stocked, and storage system tried to pick $pickedItemsQuantity"
                 }
             }
-            val quantity = Math.clamp(itemsInStockQuantity.minus(pickedItemsQuantity), 0.0, Double.MAX_VALUE)
+            val quantity = Math.clamp(itemsInStockQuantity.minus(pickedItemsQuantity).toLong(), 0, Int.MAX_VALUE)
```
MormonJesus69420, line 113: Shouldn't that be toInt instead of toLong?

MormonJesus69420 (changes requested):

src/main/kotlin/no/nb/mlt/wls/domain/WLSService.kt
anotheroneofthese: It is already an int, but the clamp function surprisingly does not take in an int.

MormonJesus69420 (approved):
LGTM

PR #49, opened 2024-11-01
MLT-0045 Add Hostname on Order IDs for SynQ Orders

anotheroneofthese:
Prevents order conflicts with multiple hosts.

Remains a draft we decide where it is best to handle the validation for this (likely domain, or should it happen in the service?).


src/main/kotlin/no/nb/mlt/wls/infrastructure/synq/SynqOrderPayload.kt
```diff
@@ -53,15 +52,15 @@ data class ShippingAddress(
 
 fun Order.toSynqPayload() =
     SynqOrderPayload(
-        orderId = hostOrderId,
+        orderId = hostName.toString().uppercase() + "_" + hostOrderId,
```
MormonJesus69420, line 55: Nooooooo... not the underscore... we hates the undersore, yes yes, we hates it... Golum! Golum!

MormonJesus69420 (approved):
Looks good, have we made a task to migrate the order logic from SynQ Controler to Order domain logic?

PR #50, opened 2024-11-05
MLT-0064 Domain improvements

anotheroneofthese:
This PR aims to improve the structure and maintability of the codebase by migrating a bunch of the domain-specific code into the domain models. Formerly these methods/functions are scattered across the service and the controllers. 

While not technically wrong, it is more concise what use-cases each domain model supports when it is collected within itself. new use-cases also naturally come up when writing this way.

### DIFF
--- a/src/main/kotlin/no/nb/mlt/wls/application/synqapi/synq/SynqOrderPickingConfirmationPayload.kt
+++ b/src/main/kotlin/no/nb/mlt/wls/application/synqapi/synq/SynqOrderPickingConfirmationPayload.kt
@@ -69,7 +69,7 @@ data class SynqOrderPickingConfirmationPayload(
 
     fun getHostname(): HostName {
         val hostString = this.orderLine.firstOrNull()?.hostName ?: throw ValidationException("Unable to get HostName as there were no order lines")
-        return HostName.valueOf(hostString)
+        return HostName.valueOf(hostString.uppercase())
     }
 
     /**
--- a/src/main/kotlin/no/nb/mlt/wls/application/synqapi/synq/SynqOrderPickingConfirmationPayload.kt
+++ b/src/main/kotlin/no/nb/mlt/wls/application/synqapi/synq/SynqOrderPickingConfirmationPayload.kt
@@ -11,7 +11,7 @@ import no.nb.mlt.wls.domain.ports.inbound.ValidationException
       "orderLine" : [
         {
           "confidentialProduct" : false,
-          "hostName" : "Axiell",
+          "hostName" : "AXIELL",
           "orderLineNumber" : 1,
           "orderTuId" : "SYS_TU_00000001157",
           "orderTuType" : "UFO",
@@ -102,7 +102,7 @@ data class SynqOrderPickingConfirmationPayload(
     example = """
     {
       "confidentialProduct" : false,
-      "hostName" : "Axiell",
+      "hostName" : "AXIELL",
       "orderLineNumber" : 1,
       "orderTuId" : "SYS_TU_00000001157",
       "orderTuType" : "UFO",
@@ -125,7 +125,7 @@ data class OrderLine(
     val confidentialProduct: Boolean,
     @Schema(
         description = "Name of the host system which placed the order/owns the order products/items.",
-        example = "Axiell"
+        example = "AXIELL"
     )
     val hostName: String,
     @Schema(
--- a/src/main/kotlin/no/nb/mlt/wls/application/synqapi/synq/SynqOrderPickingConfirmationPayload.kt
+++ b/src/main/kotlin/no/nb/mlt/wls/application/synqapi/synq/SynqOrderPickingConfirmationPayload.kt
@@ -47,14 +47,12 @@ data class SynqOrderPickingConfirmationPayload(
     )
     val warehouse: String
 ) {
-    fun validate(orderId: String) {
+    /**
+     * Validates the packet data and structure
+     */
+    fun validate() {
         if (orderLine.isEmpty()) {
             throw ValidationException("Picking update does not contain any elements in the order line")
-        }
-
-        val hostName = this.getHostname().toString()
-        if (!orderId.startsWith(hostName)) {
-            throw ValidationException("Order picked from SynQ does not have the host '$hostName' declared within WLS")
         }
 
         if (operator.isBlank()) {
@@ -66,6 +64,16 @@ data class SynqOrderPickingConfirmationPayload(
         }
 
         orderLine.forEach(OrderLine::validate)
+    }
+
+    /**
+     * Validates the order id and checks if it starts with the hostname
+     */
+    fun validateOrderId(orderId: String) {
+        val hostName = getHostname().toString()
+        if (!orderId.startsWith(hostName)) {
+            throw ValidationException("Order picked from SynQ does not have the host '$hostName' declared within WLS")
+        }
     }
 
     fun getHostname(): HostName {
--- a/src/main/kotlin/no/nb/mlt/wls/application/synqapi/synq/SynqOrderPickingConfirmationPayload.kt
+++ b/src/main/kotlin/no/nb/mlt/wls/application/synqapi/synq/SynqOrderPickingConfirmationPayload.kt
@@ -47,9 +47,13 @@ data class SynqOrderPickingConfirmationPayload(
     )
     val warehouse: String
 ) {
-    fun validate() {
+    fun validate(orderId: String) {
         if (orderLine.isEmpty()) {
             throw ValidationException("Picking update does not contain any elements in the order line")
+        }
+
+        if (!orderId.startsWith(this.getHostname().toString())) {
+            throw ValidationException("Order picked from SynQ does not have a host declared within WLS")
         }
 
         if (operator.isBlank()) {
@@ -61,6 +65,25 @@ data class SynqOrderPickingConfirmationPayload(
         }
 
         orderLine.forEach(OrderLine::validate)
+    }
+
+    fun getHostname(): HostName {
+        val hostString = this.orderLine.firstOrNull()?.hostName ?: throw ValidationException("Unable to get HostName as there were no order lines")
+        return HostName.valueOf(hostString)
+    }
+
+    /**
+     * Creates a map between the id and the quantity picked from each product in the order line
+     */
+    fun mapProductsToQuantity(): Map<String, Int> {
+        val map: MutableMap<String, Int> = mutableMapOf()
+        this.orderLine.map { orderLine ->
+            map.put(
+                orderLine.productId,
+                orderLine.quantity
+            )
+        }
+        return map
     }
 }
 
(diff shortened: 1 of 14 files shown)


src/main/kotlin/no/nb/mlt/wls/domain/model/Order.kt
```diff
@@ -29,8 +29,8 @@ data class Order(
         )
     }
 
-    fun setOrderLineStatus(
-        hostId: String,
+    private fun setOrderLineStatus(
+        hostId: List<String>,
```
MormonJesus69420, line 33: hostIds if we are having a list of them

src/main/kotlin/no/nb/mlt/wls/domain/model/Order.kt
```diff
@@ -91,11 +91,34 @@ data class Order(
...
+        return this.setOrderLines(itemIds)
+            .setCallbackUrl(callbackUrl)
+            .setOrderType(orderType)
+            .setReceiver(receiver)
+    }
+
+    /**
+     * Delete the order as long as it is possible
+     */
+    fun deleteOrder() {
```
MormonJesus69420, line 114: Wouldn't it be better to call it canDeleteOrder and have it return bool? Then use it in the service to check if order can be deleted?

src/main/kotlin/no/nb/mlt/wls/domain/model/Order.kt
```diff
@@ -91,11 +91,34 @@ data class Order(
...
+        throwIfInProgress()
+    }
+
     /**
      * Throws an exception if the order has been started or finished
-     * TODO - Make this private, and handle this in a more domain-driven way
      */
-    fun throwIfInProgress() {
+    private fun throwIfInProgress() {
```
MormonJesus69420, line 121: This is used in both delete and update, change the language of error messages to reflect that. i.e. "order can not be changed"

src/main/kotlin/no/nb/mlt/wls/domain/WLSService.kt
```diff
@@ -94,43 +86,32 @@ class WLSService(
...
-        itemIds.map { itemId ->
-            // TODO - Get All Items function?
-            val item = getItem(itemId.hostName, itemId.hostId)!!
-
+        val itemsToPick = getItems(itemsPickedMap.keys.toList(), hostName)
+        itemsToPick.map { item ->
             val itemsInStockQuantity = item.quantity ?: 0
             val pickedItemsQuantity = itemsPickedMap[item.hostId] ?: 0
 
             // In the case of over-picking, set quantity to zero
```
MormonJesus69420, line 94: Delete this comment as we set quantity in the Item::pickItem function

src/main/kotlin/no/nb/mlt/wls/domain/WLSService.kt
```diff
@@ -183,7 +154,7 @@ class WLSService(
         hostOrderId: String
     ) {
         val order = getOrderOrThrow(hostName, hostOrderId)
-        order.throwIfInProgress()
+        order.deleteOrder()
```
MormonJesus69420, line 157: I think I already mentioned it, but a change of name would be good here, or change how this is handled

src/main/kotlin/no/nb/mlt/wls/infrastructure/repositories/item/MongoItemRepository.kt
```diff
@@ -107,4 +123,9 @@ interface ItemMongoRepository : ReactiveMongoRepository<MongoItem, String> {
         quantity: Int,
         location: String
     ): Mono<Long>
+
+    fun findByHostNameAndHostId(
```
MormonJesus69420, line 127: findAllByHostNameAndHostId

src/test/kotlin/no/nb/mlt/wls/synq/model/SynqModelValidationTest.kt
```diff
@@ -35,14 +35,33 @@ class SynqModelValidationTest {
 
     @Test
     fun `valid SynqOrderPickingConfirmationPayload should pass validation`() {
-        thenCode(validSynqOrderPickingConfirmationPayload::validate).doesNotThrowAnyException()
+        thenCode {
+            validSynqOrderPickingConfirmationPayload.validate()
```
MormonJesus69420, line 39: Maybe go back to the old test calls since validate doesn't take any input anymore

MormonJesus69420 (changes requested):

src/main/kotlin/no/nb/mlt/wls/domain/WLSService.kt
anotheroneofthese: Should I move the logging to the Item domain class?

MormonJesus69420: I think logging can stay here, especially since this is more issue of operation than the model. It's not the item's fault that the quantity is fucky wucky

MormonJesus69420: Yeah you're right, especially since we already log in the item class, and technically it is something item/domain is more concerned about

anotheroneofthese: As discussed this is an operation that should happen in the model, as the item describes how it should behave, not the controller

src/main/kotlin/no/nb/mlt/wls/application/synqapi/synq/SynqOrderPickingConfirmationPayload.kt
```diff
@@ -59,17 +62,47 @@ data class SynqOrderPickingConfirmationPayload(
         if (warehouse.isBlank()) {
             throw ValidationException("Picking update's warehouse can not be blank")
         }
+        // Validates the hostname present on the order lines
```
MormonJesus69420, line 65:
```suggestion

        // Validates the hostname based on the value of hostname in order lines
```

src/test/kotlin/no/nb/mlt/wls/synq/model/SynqModelValidationTest.kt
```diff
@@ -35,14 +35,17 @@ class SynqModelValidationTest {
...
     }
 
     @Test
     fun `SynqOrderPickingConfirmationPayload with no order line should fail validation`() {
         val payload = validSynqOrderPickingConfirmationPayload.copy(orderLine = emptyList())
 
-        val thrown = catchThrowable(payload::validate)
+        val thrown =
+            catchThrowable {
+                payload.validate()
```
MormonJesus69420, line 47: This and all the other methods testing validate can be changed back

MormonJesus69420 (changes requested):

src/test/kotlin/no/nb/mlt/wls/synq/model/SynqModelValidationTest.kt
```diff
@@ -35,14 +35,15 @@ class SynqModelValidationTest {
...
+        thenCode(validSynqOrderStatusUpdatePayload::validate).doesNotThrowAnyException()
     }
 
     @Test
     fun `SynqOrderPickingConfirmationPayload with no order line should fail validation`() {
         val payload = validSynqOrderPickingConfirmationPayload.copy(orderLine = emptyList())
 
-        val thrown = catchThrowable(payload::validate)
+        val thrown =
+            catchThrowable(payload::validate)
```
MormonJesus69420, line 46:
```suggestion
        val thrown = catchThrowable(payload::validate)
```

src/test/kotlin/no/nb/mlt/wls/synq/model/SynqModelValidationTest.kt
```diff
@@ -54,7 +55,8 @@ class SynqModelValidationTest {
     fun `SynqOrderPickingConfirmationPayload with blank operator should fail validation`() {
         val payload = validSynqOrderPickingConfirmationPayload.copy(operator = "")
 
-        val thrown = catchThrowable(payload::validate)
+        val thrown =
+            catchThrowable(payload::validate)
```
MormonJesus69420, line 59:
```suggestion
        val thrown = catchThrowable(payload::validate)
```

src/test/kotlin/no/nb/mlt/wls/synq/model/SynqModelValidationTest.kt
```diff
@@ -66,7 +68,8 @@ class SynqModelValidationTest {
     fun `SynqOrderPickingConfirmationPayload with blank warehouse should fail validation`() {
         val payload = validSynqOrderPickingConfirmationPayload.copy(warehouse = "")
 
-        val thrown = catchThrowable(payload::validate)
+        val thrown =
+            catchThrowable(payload::validate)
```
MormonJesus69420, line 72:
```suggestion
        val thrown = catchThrowable(payload::validate)
```

src/test/kotlin/no/nb/mlt/wls/synq/model/SynqModelValidationTest.kt
```diff
@@ -84,7 +87,8 @@ class SynqModelValidationTest {
                     )
             )
 
-        val thrown = catchThrowable(payload::validate)
+        val thrown =
+            catchThrowable(payload::validate)
```
MormonJesus69420, line 91:
```suggestion
        val thrown = catchThrowable(payload::validate)
```

src/main/kotlin/no/nb/mlt/wls/domain/model/Item.kt
```diff
@@ -11,4 +15,32 @@ data class Item(
...
+            if (quantity == 0) {
+                "WITH_LENDER"
+            } else if (location != null) {
+                location
+            } else {
+                // Rare edge case. Log it until we can determine if this actually happens in production
+                logger.error {
+                    "Item with ID $hostId for host $hostName without a location was picked. Location was set to empty string."
+                }
+                ""
```
MormonJesus69420, line 42: Maybe set this to something more meaningful, like "UNKNOWN"? Since "" can happen by accident, whereas if location is "UNKNOWN" then we are both sure that it was intentionally set, and where it came from, instead of it being a mistake that someone left out an empty string.
I might also be misremembering, but I think Item's validation function throws if location is set to an empty string? Or at least I feel like it should.

MormonJesus69420 (changes requested):

src/main/kotlin/no/nb/mlt/wls/domain/model/Item.kt
MormonJesus69420:
```suggestion
                "UNKNOWN"
```

MormonJesus69420, line 42:
```suggestion
                    "Item with ID $hostId for host $hostName without a location was picked. Location was set to \"UNKNOWN\"."
                }
                "UNKNOWN"
```

MormonJesus69420 (approved):

PR #52, opened 2024-11-28
MLT-0075 Order Domain changes

anotheroneofthese:
Replaces Receiver with two new fields in the Order model

### DIFF
--- a/src/main/kotlin/no/nb/mlt/wls/domain/WLSService.kt
+++ b/src/main/kotlin/no/nb/mlt/wls/domain/WLSService.kt
@@ -164,7 +164,14 @@ class WLSService(
 
         val order = getOrderOrThrow(hostName, hostOrderId)
 
-        val updatedOrder = order.updateOrder(itemHostIds, callbackUrl, orderType, contactPerson)
+        val updatedOrder =
+            order.updateOrder(
+                itemHostIds,
+                callbackUrl,
+                orderType,
+                address ?: order.address,
+                contactPerson
+            )
         val result = storageSystemFacade.updateOrder(updatedOrder)
         return orderRepository.updateOrder(result)
     }
--- a/src/main/kotlin/no/nb/mlt/wls/domain/WLSService.kt
+++ b/src/main/kotlin/no/nb/mlt/wls/domain/WLSService.kt
@@ -153,7 +153,8 @@ class WLSService(
         hostOrderId: String,
         itemHostIds: List<String>,
         orderType: Order.Type,
-        receiver: Order.Receiver,
+        contactPerson: String,
+        address: Order.Address?,
         callbackUrl: String
     ): Order {
         val itemIds = itemHostIds.map { ItemId(hostName, it) }
@@ -163,7 +164,7 @@ class WLSService(
 
         val order = getOrderOrThrow(hostName, hostOrderId)
 
-        val updatedOrder = order.updateOrder(itemHostIds, callbackUrl, orderType, receiver)
+        val updatedOrder = order.updateOrder(itemHostIds, callbackUrl, orderType, contactPerson)
         val result = storageSystemFacade.updateOrder(updatedOrder)
         return orderRepository.updateOrder(result)
     }
--- a/src/main/kotlin/no/nb/mlt/wls/domain/WLSService.kt
+++ b/src/main/kotlin/no/nb/mlt/wls/domain/WLSService.kt
@@ -155,6 +155,7 @@ class WLSService(
         orderType: Order.Type,
         contactPerson: String,
         address: Order.Address?,
+        note: String?,
         callbackUrl: String
     ): Order {
         val itemIds = itemHostIds.map { ItemId(hostName, it) }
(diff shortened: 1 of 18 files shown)


src/main/kotlin/no/nb/mlt/wls/domain/WLSService.kt
```diff
@@ -163,7 +164,7 @@ class WLSService(
 
         val order = getOrderOrThrow(hostName, hostOrderId)
 
-        val updatedOrder = order.updateOrder(itemHostIds, callbackUrl, orderType, receiver)
+        val updatedOrder = order.updateOrder(itemHostIds, callbackUrl, orderType, contactPerson)
```
MormonJesus69420, line 167: Should be able to edit address too, no?

src/main/kotlin/no/nb/mlt/wls/domain/model/Order.kt
```diff
@@ -98,14 +99,14 @@ data class Order(
         itemIds: List<String>,
         callbackUrl: String,
         orderType: Type,
-        receiver: Receiver
+        contactPerson: String
```
MormonJesus69420, line 102: Same comment here, address should be editable

src/main/kotlin/no/nb/mlt/wls/domain/model/Order.kt
```diff
@@ -153,10 +154,37 @@ data class Order(
...
+        val name: String?,
+        val addressLine1: String?,
+        val addressLine2: String?,
+        val zipcode: String?,
+        val city: String?,
+        val state: String?,
+        val country: String?
+    ) {
+        fun validate() {
+            if (name?.isBlank() == true) {
```
MormonJesus69420, line 167: == true is redundant

src/main/kotlin/no/nb/mlt/wls/infrastructure/synq/SynqOrderPayload.kt
```diff
@@ -40,6 +40,7 @@ data class SynqOrderPayload(
     }
 }
 
+// TODO - Should this be expanded with more fields?
```
MormonJesus69420, line 43: Yes

MormonJesus69420 (changes requested):

src/main/kotlin/no/nb/mlt/wls/domain/model/Order.kt
anotheroneofthese: Nullability is weird

src/main/kotlin/no/nb/mlt/wls/infrastructure/callbacks/NotificationOrderPayload.kt
```diff
@@ -59,9 +65,13 @@ data class NotificationOrderPayload(
...
     @Schema(
-        description = "Who's the receiver of the material in the order."
+        description = "The person to contact regarding matters about the order"
     )
-    val receiver: Order.Receiver,
+    val contactPerson: String,
+    @Schema(
+        description = "The address of the receiver of the material in the order."
+    )
+    val address: Order.Address?,
```
MormonJesus69420, line 74: Notice: note not noted in notification, not nice, no, no... Note the note in notification now.

src/main/kotlin/no/nb/mlt/wls/infrastructure/synq/SynqOrderPayload.kt
```diff
@@ -43,10 +43,16 @@ data class SynqOrderPayload(
 data class ShippingAddress(
     val address: Address
 ) {
+    // TODO - Should all the fields here be annotated?
```
MormonJesus69420, line 46: Yeah, I think the address as a whole and its fields should not be included if null

MormonJesus69420 (changes requested):

src/main/kotlin/no/nb/mlt/wls/infrastructure/synq/SynqOrderPayload.kt
anotheroneofthese: The ShippingAddress is currently not nullable, due to `contactPerson` being a required field for the SynQ payload. Should this information be sent to storage? If not, then we can make it an optional field

MormonJesus69420:
I dunno if you and Tom have discussed it, but shouldn't the  person in the address be nullable as well?

MormonJesus69420 (approved):
Looks good, why are ports in compose file written in quotes now?

EDIT: oh okay so apparently anything in YAML written as xx:xx will be parsed as a base 60 number if values are under 60, weird but okay

anotheroneofthese:
New best practice, according to https://stackoverflow.com/questions/58810789/quotes-on-docker-compose-yml-ports-make-any-difference
image
Docker doesn't seem to reflect this in their specification anymore. They might have changed it, although there seems to be questions floating around where the solution usually is stringifying the ports

MormonJesus69420:
image
https://yaml.org/type/float.html

Yeah it is still recommended practice to write HOST:CONTAINER as strings

PR #55, opened 2024-12-18
Host API Controller changes

anotheroneofthese:
This PR removes the unused status field from the order payload when creating orders (MLT-0076). 
It does this by introducing a new payload, `ApiCreateOrderPayload`, which only accepts fields we will actually process (MLT-0083). The latter is also applied for items, via `ApiCreateItemPayload`

### DIFF
--- a/src/main/kotlin/no/nb/mlt/wls/application/hostapi/item/ApiCreateItemPayload.kt
+++ b/src/main/kotlin/no/nb/mlt/wls/application/hostapi/item/ApiCreateItemPayload.kt
@@ -1,0 +1,130 @@
+package no.nb.mlt.wls.application.hostapi.item
+
+import io.swagger.v3.oas.annotations.media.Schema
+import no.nb.mlt.wls.domain.model.Environment
+import no.nb.mlt.wls.domain.model.HostName
+import no.nb.mlt.wls.domain.model.Item
+import no.nb.mlt.wls.domain.model.ItemCategory
+import no.nb.mlt.wls.domain.model.Packaging
+import no.nb.mlt.wls.domain.ports.inbound.ItemMetadata
+import no.nb.mlt.wls.domain.ports.inbound.ValidationException
+import org.apache.commons.validator.routines.UrlValidator
+
+@Schema(
+    description = """Payload for registering an item in Hermes WLS, and appropriate storage systems.""",
+    example = """
+    {
+      "hostId": "mlt-12345",
+      "hostName": "AXIELL",
+      "description": "Tyven, tyven skal du hete",
+      "itemCategory": "PAPER",
+      "preferredEnvironment": "NONE",
+      "packaging": "NONE",
+      "callbackUrl": "https://callback-wls.no/item"
+    }
+    """
+)
+data class ApiCreateItemPayload(
+    @Schema(
+        description = """The item ID from the host system, usually a barcode or an equivalent ID.""",
+        example = "mlt-12345"
+    )
+    val hostId: String,
+    @Schema(
+        description = """Name of the host system which the item originates from.
+                Host system is usually the catalogue that the item is registered in.""",
+        examples = [ "AXIELL", "ALMA", "ASTA", "BIBLIOFIL" ]
+    )
+    val hostName: HostName,
+    @Schema(
+        description = """Description of the item for easy identification in the warehouse system.
+                Usually an item title/name, e.g. book title, film name, etc. or contents description.""",
+        examples = ["Tyven, tyven skal du hete", "Avisa Hemnes", "Kill Buljo"]
+    )
+    val description: String,
+    @Schema(
+        description = """Item's category, same category indicates that the items can be stored together without any preservation issues.
+                For example: books, magazines, newspapers, etc. are of type PAPER, and can be stored together without damaging each other.""",
+        examples = ["PAPER", "DISC", "FILM", "PHOTO", "EQUIPMENT", "BULK_ITEMS", "MAGNETIC_TAPE"]
+    )
+    val itemCategory: ItemCategory,
+    @Schema(
+        description = """What kind of environment the item should be stored in.
+                "NONE" is for normal storage for the item category, "FRYS" is for frozen storage, etc.
+                NOTE: This is not a guarantee that the item will be stored in the preferred environment.
+                In cases where storage space is limited, the item may be stored in a different environment.""",
+        examples = ["NONE", "FRYS"]
+    )
+    val preferredEnvironment: Environment,
+    @Schema(
+        description = """Whether the item is a single object or a container with other items inside.
+                "NONE" is for single objects, "ABOX" is for archival boxes, etc.
+                NOTE: It is up to the catalogue to keep track of the items inside a container.""",
+        examples = ["NONE", "BOX", "ABOX", "CRATE"]
+    )
+    val packaging: Packaging,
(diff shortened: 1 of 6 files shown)


src/main/kotlin/no/nb/mlt/wls/application/hostapi/item/ApiCreateItemPayload.kt
```diff
@@ -0,0 +1,130 @@
...
+        Item(
+            hostId = hostId,
+            hostName = hostName,
+            description = description,
+            itemCategory = itemCategory,
+            preferredEnvironment = preferredEnvironment,
+            packaging = packaging,
+            callbackUrl = callbackUrl,
+            location = null,
+            quantity = null
```
MormonJesus69420, line 83: Quantity can be set to 0, as it helps with checks and not having to worry about NPE

src/main/kotlin/no/nb/mlt/wls/application/hostapi/order/ApiCreateOrderPayload.kt
```diff
@@ -0,0 +1,137 @@
...
+import no.nb.mlt.wls.domain.ports.inbound.ValidationException
+import org.apache.commons.validator.routines.UrlValidator
+
+@Schema(
+    description = """Payload for creating orders in Hermes WLS, and appropriate storage system(s).""",
+    example = """
+    {
+      "hostName": "AXIELL",
+      "hostOrderId": "mlt-12345-order",
+      "status": "NOT_STARTED",
```
MormonJesus69420, line 17: Yeet

src/main/kotlin/no/nb/mlt/wls/application/hostapi/order/ApiCreateOrderPayload.kt
```diff
@@ -0,0 +1,137 @@
...
+        // Yes I am aware that this function is duplicated in three places
+        // But I prefer readability over DRY in cases like this
+
+        val validator = UrlValidator(arrayOf("http", "https")) // Allow only HTTP/HTTPS
+        return validator.isValid(url)
+    }
+}
+
+fun Order.toCreateApiOrderPayload() =
+    ApiOrderPayload(
```
MormonJesus69420, line 127: Wrong class

src/main/kotlin/no/nb/mlt/wls/application/hostapi/order/ApiCreateOrderPayload.kt
```diff
@@ -0,0 +1,137 @@
...
+        hostName = hostName,
+        hostOrderId = hostOrderId,
+        status = status,
```
MormonJesus69420, line 130: Should be removed (i mean compiler won't accept it if we change to proper class above.

src/main/kotlin/no/nb/mlt/wls/application/hostapi/item/ApiCreateItemPayload.kt
```diff
@@ -0,0 +1,130 @@
...
+    fun toItem(): Item =
+        Item(
+            hostId = hostId,
+            hostName = hostName,
+            description = description,
+            itemCategory = itemCategory,
+            preferredEnvironment = preferredEnvironment,
+            packaging = packaging,
+            callbackUrl = callbackUrl,
+            location = null,
```
MormonJesus69420, line 82: Up to debate, but how about we set this to `UNKNOWN` or similar?
Potential benefits:
- Clear definition where item is located (as far as we are aware)
- Can make both location and quantity non-nullable in the Item model (less null value handling)
- Makes value explicit (avoids confusion as to whether null is intentional or not)

Potential drawbacks:
- User/Host must be aware that UNKNOWN is a possible value and can handle it properly.
    - This is only an issue when creating item, after it has gone into the system it should never have value NULL for location.

MormonJesus69420 (changes requested):
**_NOTE_**:

> I think that in ApiItemPayload the fields location and quantity can get rid of their 
>     `accessMode = READ_ONLY`
> schema attribute

src/main/kotlin/no/nb/mlt/wls/application/hostapi/item/ApiCreateItemPayload.kt
MormonJesus69420: Yeah it goes back to the other comment I made about getting rid of nulls and just setting default values. If we don't do it, then we still have to deal with nulls

src/main/kotlin/no/nb/mlt/wls/application/hostapi/item/ApiCreateItemPayload.kt
anotheroneofthese: Will write a task up for changing nullability, which should be handled in a following PR :+1:

src/main/kotlin/no/nb/mlt/wls/application/hostapi/item/ApiItemPayload.kt
```diff
@@ -73,15 +72,13 @@ data class ApiItemPayload(
     @Schema(
         description = """Where the item is located, can be used for tracking item movement through storage systems.""",
         examples = ["SYNQ_WAREHOUSE", "AUTOSTORE", "KARDEX"],
-        accessMode = READ_ONLY,
         required = false
     )
     val location: String?,
```
MormonJesus69420, line 77: Can probably mark these as non-nullable
```suggestion
    val location: String,
```

src/main/kotlin/no/nb/mlt/wls/application/hostapi/item/ApiItemPayload.kt
```diff
@@ -73,15 +72,13 @@ data class ApiItemPayload(
...
     @Schema(
         description = """Quantity on hand of the item, this easily denotes if the item is in the storage or not.
                 If the item is in storage then quantity is 1, if it's not in storage then quantity is 0.""",
         examples = [ "0", "1"],
-        accessMode = READ_ONLY,
         required = false
     )
     val quantity: Int?
```
MormonJesus69420, line 84: Can probably mark these as non-nullable
```suggestion
    val quantity: Int
```

src/main/kotlin/no/nb/mlt/wls/application/hostapi/item/ApiItemPayload.kt
```diff
@@ -73,15 +72,13 @@ data class ApiItemPayload(
     @Schema(
         description = """Where the item is located, can be used for tracking item movement through storage systems.""",
         examples = ["SYNQ_WAREHOUSE", "AUTOSTORE", "KARDEX"],
```
MormonJesus69420, line 74: Mayhaps add "UNKNOWN" as a valid example
```suggestion
        examples = ["UNKNOWN", "SYNQ_WAREHOUSE", "AUTOSTORE", "KARDEX"],
```

src/main/kotlin/no/nb/mlt/wls/application/hostapi/item/ApiItemPayload.kt
MormonJesus69420: Should also do the same in the notification payload

MormonJesus69420 (changes requested):

MormonJesus69420 (approved):

PR #56, opened 2024-12-19
MLT-0086 Item Model Update

anotheroneofthese:
Follows up on #55, and makes the `quantity` and `location` fields within the data non-nullable. This comes with more benefits than downsides, and simplifies the API.

It even removes the strange edge-case of picking an item where the location is null.  
```diff
        } else if (location != null) {
            location
        } else {
-            // Rare edge case. Log it until we can determine if this actually happens in production
-            logger.error {
-                "Item with ID '$hostId' for host '$hostName' without a location was picked. Location was set to 'UNKNOWN'."
-            }
            "UNKNOWN"
        }
```

### DIFF
--- a/src/main/kotlin/no/nb/mlt/wls/application/hostapi/item/ApiItemPayload.kt
+++ b/src/main/kotlin/no/nb/mlt/wls/application/hostapi/item/ApiItemPayload.kt
@@ -74,14 +74,13 @@ data class ApiItemPayload(
         examples = ["UNKNOWN", "WITH_LENDER", "SYNQ_WAREHOUSE", "AUTOSTORE", "KARDEX"],
         required = false
     )
-    val location: String?,
+    val location: String,
     @Schema(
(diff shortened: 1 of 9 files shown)


src/main/kotlin/no/nb/mlt/wls/application/hostapi/item/ApiItemPayload.kt
```diff
@@ -74,14 +73,13 @@ data class ApiItemPayload(
         examples = ["UNKNOWN", "WITH_LENDER", "SYNQ_WAREHOUSE", "AUTOSTORE", "KARDEX"],
         required = false
```
MormonJesus69420, line 74: This can be removed

MormonJesus69420 (changes requested):
Just a small nitpick, otherwise accepted

MormonJesus69420 (approved):

PR #58, opened 2025-01-06
MLT-0072 Emails for Orders

anotheroneofthese:
Email handling for orders


pom.xml
```diff
@@ -170,6 +183,17 @@
...
 
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-starter-freemarker</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>io.nayuki</groupId>
+      <artifactId>qrcodegen</artifactId>
+      <version>1.8.0</version>
```
MormonJesus69420, line 194: Pwopewty pwease `${qrcodegen.version}`

src/main/kotlin/no/nb/mlt/wls/domain/WLSService.kt
```diff
@@ -135,6 +137,12 @@ class WLSService(
         }
 
         val order = orderRepository.createOrder(orderDTO.toOrder())
+
+        // email handling
+        val items = order.orderLine.map { it.hostId }
+        val orderItems = getItems(items, order.hostName)
+        if (orderItems.isNotEmpty()) emailAdapter.orderCreated(order, orderItems)
```
MormonJesus69420, line 144: Feel like this could be in its own little function for better clarity

src/main/kotlin/no/nb/mlt/wls/infrastructure/email/BarcodeUtils.kt
```diff
@@ -0,0 +1,81 @@
...
+
+            for (y in 0 until size) {
+                for (x in 0 until size) {
+                    val isDark = qr.getModule(x / scale - border, y / scale - border)
+                    val color =
+                        if (isDark) {
+                            0xFFFFE0
+                        } else {
+                            0xAE4000
+                        }
```
MormonJesus69420, line 67:
```suggestion
                    val color = if (isDark) 0xFFFFE0 else 0xAE4000
```
Also why do you set piss yellow when isDark is true, and red if it's false? I think it's safer to just go with standard black and white

src/main/resources/templates/email/order-confirmation.ftl
```diff
@@ -0,0 +1,23 @@
+<div>
+    Bestillingen din har blitt mottatt.
+    <p>
```
MormonJesus69420, line 3: Mayhaps instead of all those pp's we could use description list?
image

src/main/resources/application.yml
```diff
@@ -73,11 +80,20 @@ spring:
...
+    port: ${EMAIL_PORT}
+
 # Set a dummycatalog base path to SynQ for testing purposes
 synq:
   path:
     base: http://localhost:8181/synq/resources
 
+# Configures destination e-mails for receivers
+mail:
+  storage: ${EMAIL_STORAGE}
```
MormonJesus69420, line 95: Please, I beg, call this rolodex instead of email_storage

src/main/kotlin/no/nb/mlt/wls/infrastructure/email/EmailAdapter.kt
```diff
@@ -0,0 +1,159 @@
...
+import java.awt.image.BufferedImage
+
+private val logger = KotlinLogging.logger {}
+
+class EmailAdapter(
+    private val emailRepository: EmailRepository,
+    private val emailSender: JavaMailSender,
+    private val freeMarkerConfigurer: FreeMarkerConfigurer
+) : EmailNotifier {
+    @Value("\${mail.storage}")
```
MormonJesus69420, line 26: Maybe something like "warehouse.email.address"

Might also consider distancing ourselves from host and warehouse concepts here. I would maybe consider names like order handler and order creator or something like that

src/main/kotlin/no/nb/mlt/wls/infrastructure/email/EmailAdapter.kt
```diff
@@ -0,0 +1,159 @@
...
+    ) {
+        val receiver = emailRepository.getHostEmail(order.hostName) ?: return
+        try {
+            sendOrderEmail(createOrderEmail(order, receiver))
+            sendHostOrderEmail(createStorageOrderEmail(order, orderItems))
+            logger.info {
+                "Emails sent for order ${order.hostOrderId}"
+            }
+        } catch (e: MailException) {
+            // TODO - Review. Should probably be explicitly logged?
```
MormonJesus69420, line 41: Yeah,just log some simple info, especially in prod, print stack trace in stage only?

src/main/kotlin/no/nb/mlt/wls/infrastructure/email/EmailAdapter.kt
```diff
@@ -0,0 +1,159 @@
...
+        val type = "order-confirmation.ftl"
+        val mail = emailSender.createMimeMessage()
+        val helper = MimeMessageHelper(mail, false)
+        val template = freeMarkerConfigurer.configuration.getTemplate(type)
+        val htmlBody = FreeMarkerTemplateUtils.processTemplateIntoString(template, mapOf("order" to order))
+
+        // Email Metadata
+        helper.setText(htmlBody, true)
+        helper.setSubject("Bestillingsbekreftelse fra WLS - ${order.hostOrderId}")
+        helper.setFrom("noreply@nb.no")
```
MormonJesus69420, line 91: might be best to not include our website here. Don't want to risk getting nb.no marked as spam sender

src/main/kotlin/no/nb/mlt/wls/infrastructure/email/EmailAdapter.kt
```diff
@@ -0,0 +1,159 @@
...
+        val list = mutableListOf<EmailOrderItem>()
+        for (item in items) {
+            list.add(EmailOrderItem(item, getQrHtmlString(item.hostId), BarcodeUtils.createQrImage(item.hostId)))
+        }
+        return list
+    }
+
+    private fun getQrHtmlString(cid: String): String =
+        """
+            <img src="cid:qr-$cid" alt="qrcode"/></img>
```
MormonJesus69420, line 155: wouldn't it be better idea to set the alt tag with orderID or itemID so you can at least get the ID in text if QR code won't display?

MormonJesus69420 (changes requested):

src/main/kotlin/no/nb/mlt/wls/infrastructure/email/EmailAdapter.kt
anotheroneofthese: This is possible. I don't see it as strictly necessary as the template does put captions beneath the images anyway. You would end up with something like
```
| "QR of mlt-12345" |
| "mlt-12345"       |
```
per-item.

src/main/resources/templates/email/order-confirmation.ftl
anotheroneofthese: The intent behind the paragraph tags was to try minimize using a ton of divs. Regardless of what is used, it produces the same output

src/main/kotlin/no/nb/mlt/wls/infrastructure/email/EmailAdapter.kt
anotheroneofthese: Is checking for debug sufficient, or does this need some more explicit/specific check?

MormonJesus69420 (approved):

PR #59, opened 2025-01-07
MLT-0080 Change authorization to roles

MormonJesus69420:
This PR attempts to change the authorization model to use Roles instead of Scopes as advised by Tom. Changes the JWT algorithm to ES256 as recommended by the documentation.

Questions to @tomolse and @anotheroneofthese: 
- Should we validate JWT type? When I was reading up on security with JWT they recommend checking token type is valid, in our case Bearer and not something else.
- How long should we set JWT time/expiry to? Mostly thinking in local setup at the moment, but I think this should be the same in every environment, local, stage, and prod.

### DIFF
--- a/docker/keycloak/import/mlt-local-realm-export.json
+++ b/docker/keycloak/import/mlt-local-realm-export.json
@@ -56,6 +56,22 @@
       "containerId" : "c87b4b7b-5d42-4304-aec1-68b6817a7025",
       "attributes" : { }
     }, {
+      "id" : "a4e99e98-ede2-4cb6-8ec8-344f584443ac",
+      "name" : "order",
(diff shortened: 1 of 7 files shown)


tomolse:
> * Should we validate JWT type? When I was reading up on security with JWT they recommend checking token type is valid, in our case Bearer and not something else.

I think the standard JWT validation in Spring is enough

> * How long should we set JWT time/expiry to? Mostly thinking in local setup at the moment, but I think this should be the same in every environment, local, stage, and prod.

The expiration should be as short as possible because its hard to revoke. Maybe start with something between 1-2 minutes.

tomolse (approved):

PR #61, opened 2025-01-31
Update OrderLine Swagger examples

anotheroneofthese:
Excludes the status field when creating orders as this is not required

### DIFF
--- a/src/main/kotlin/no/nb/mlt/wls/application/hostapi/order/ApiOrderPayload.kt
+++ b/src/main/kotlin/no/nb/mlt/wls/application/hostapi/order/ApiOrderPayload.kt
@@ -160,7 +160,7 @@ data class OrderLine(
     val hostId: String,
     @Schema(
         description = """Current status for the ordered item.""",
-        examples = ["NOT_STARTED", "PICKED", "FAILED"],
+        examples = ["NOT_STARTED", "PICKED", "RETURNED", "FAILED"],
         accessMode = READ_ONLY
     )
     val status: Order.OrderItem.Status?
--- a/src/main/kotlin/no/nb/mlt/wls/application/hostapi/order/ApiOrderPayload.kt
+++ b/src/main/kotlin/no/nb/mlt/wls/application/hostapi/order/ApiOrderPayload.kt
@@ -160,7 +160,8 @@ data class OrderLine(
     val hostId: String,
     @Schema(
         description = """Current status for the ordered item.""",
-        examples = ["NOT_STARTED", "PICKED", "FAILED"]
+        examples = ["NOT_STARTED", "PICKED", "FAILED"],
+        accessMode = READ_ONLY
     )
     val status: Order.OrderItem.Status?
 ) {
--- a/src/main/kotlin/no/nb/mlt/wls/domain/model/Order.kt
+++ b/src/main/kotlin/no/nb/mlt/wls/domain/model/Order.kt
@@ -173,6 +173,27 @@ data class Order(
             FAILED,
             RETURNED
         }
+
+        fun isPickedOrFailed(): Boolean {
+            return when (this.status) {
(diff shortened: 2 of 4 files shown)


src/main/kotlin/no/nb/mlt/wls/application/hostapi/order/ApiOrderPayload.kt
```diff
@@ -160,7 +160,8 @@ data class OrderLine(
     val hostId: String,
     @Schema(
         description = """Current status for the ordered item.""",
-        examples = ["NOT_STARTED", "PICKED", "FAILED"]
+        examples = ["NOT_STARTED", "PICKED", "FAILED"],
```
MormonJesus69420, line 163:
```suggestion
        examples = ["NOT_STARTED", "PICKED", "RETURNED", "FAILED"],
```

src/main/kotlin/no/nb/mlt/wls/domain/model/Order.kt
```diff
@@ -72,7 +71,16 @@ data class Order(
     }
 
     private fun updateOrderStatusFromOrderLines(): Order {
+        // This might benefit from a small refactor of sorts
         return when {
+            orderLine.all { it.status == OrderItem.Status.RETURNED } -> {
```
MormonJesus69420, line 76: Might be an idea to make a method for it like in other cases, both for consistency and to make it easier to change conditions for returned objects in the future.

src/main/kotlin/no/nb/mlt/wls/domain/model/Order.kt
```diff
@@ -213,8 +223,25 @@ data class Order(
     }
 }
 
-fun Order.OrderItem.isPickedOrFailed(): Boolean {
-    return this.status == PICKED || this.status == FAILED
+fun OrderItem.isPickedOrFailed(): Boolean {
```
MormonJesus69420, line 226: Can't these just be class functions, I don't see a need to make them extension functions in this case

src/main/kotlin/no/nb/mlt/wls/domain/model/Order.kt
```diff
@@ -213,8 +223,25 @@ data class Order(
...
+    return when (this.status) {
+        OrderItem.Status.NOT_STARTED -> false
+        OrderItem.Status.PICKED -> true
+        OrderItem.Status.FAILED -> true
+        OrderItem.Status.RETURNED -> false
```
MormonJesus69420, line 231: IIRC if we bump our Kotlin to 2.1.0 we can make the `when` a bit cleaner by being able to drop the `OrderItem.Status.` prefix. Though I am not sure where I read that.

MormonJesus69420 (changes requested):

src/main/kotlin/no/nb/mlt/wls/domain/model/Order.kt
anotheroneofthese: The prefix exists for clarity - the same status exists for the main `Order` class as well, and as such I would like it to be clear which one you are using. The used to be a static import actually, which was removed with this PR

src/main/kotlin/no/nb/mlt/wls/domain/model/Order.kt
```diff
@@ -35,7 +33,7 @@ data class Order(
         hostIds: List<String>,
         status: OrderItem.Status
     ): Order {
-        if (isOrderClosed()) {
+        if (isOrderClosed() && status != OrderItem.Status.RETURNED) {
```
MormonJesus69420, line 36: Wait, shouldn't this be using the Order's Status enum, or am I tripping?

MormonJesus69420 (approved):
I think this looks good

PR #62, opened 2025-02-06
MLT-77: Add signature and timestamp header to callback on order and item

tomolse:


src/main/resources/application.yml
```diff
@@ -73,6 +73,10 @@ spring:
       password: drossap
       database: wls
 
+# Secret for the signature on callbacks for the local-dev profile
+callback:
+  secret: superdupersecretkey
```
MormonJesus69420, line 78: Secrets as important as this one should contain at least one uppercase letter, a number, and a special character. I propose to change it to `SuperDuper$ecretKey69420` for added security

src/main/resources/application.yml
```diff
@@ -108,6 +112,9 @@ synq:
   path:
     base: ${SYNQ_BASE_URL}
 
+# Secret for the signature on callbacks in stage environment
+callback:
+  secret: ${CALLBACK_SECRET}
```
MormonJesus69420, line 117: I prefer to have some room in the sourcecode. 
```suggestion
  secret: ${CALLBACK_SECRET}

```

MormonJesus69420 (commented):
Looks good to me, I would like to have a talk about the `generateSignature` func from InventoryNotifierAdapter class so I can understand it better. As some of it seems a bit esoteric.

MormonJesus69420 (approved):

pom.xml
```diff
@@ -218,6 +218,12 @@
       <version>${archunit.junit5.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.squareup.okhttp3</groupId>
+      <artifactId>mockwebserver</artifactId>
+      <version>4.12.0</version>
```
MormonJesus69420, line 224: Oh right, shouldn't this be a property, ala `${mockwebserver.version}`?

MormonJesus69420 (changes requested):

src/main/kotlin/no/nb/mlt/wls/infrastructure/callbacks/InventoryNotifierAdapter.kt
```diff
@@ -1,21 +1,35 @@
...
+        val signature = generateSignature(payload, timestamp)
+
         if (item.callbackUrl != null) {
             webClient
                 .post()
                 .uri(item.callbackUrl)
-                .bodyValue(item.toNotificationItemPayload())
+                .bodyValue(payload)
+                .header("X-Signature", signature)
+                .header("X-Timestamp", timestamp)
```
anotheroneofthese, line 32: When testing this locally, I had issues getting the request to parse correctly. This was a side-effect of the request not declaring its content type, which is something we should do:

```suggestion
                .header("X-Timestamp", timestamp)
                .header(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON_VALUE)
```

src/main/kotlin/no/nb/mlt/wls/infrastructure/callbacks/InventoryNotifierAdapter.kt
```diff
@@ -25,15 +39,35 @@ class InventoryNotifierAdapter(
...
+        val signature = generateSignature(payload, timestamp)
+
         // TODO: Should probably have a more robust retry mechanism, what if receiver is down for a while?
         webClient
             .post()
             .uri(order.callbackUrl)
-            .bodyValue(order.toNotificationOrderPayload())
+            .bodyValue(payload)
+            .header("X-Signature", signature)
+            .header("X-Timestamp", timestamp)
```
anotheroneofthese, line 52: See similar comment above

```suggestion
            .header("X-Timestamp", timestamp)
            .header(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON_VALUE)
```

anotheroneofthese (dismissed):

PR #69, opened 2025-03-03
MLT-0117 Use separate web clients for proxy and non-proxy hosts

MormonJesus69420:
Works just fine locally, but the problem is that I can't really test it in `InventoryNotifierAdapterTest`


anotheroneofthese (approved):
Really cool seeing a SpyK definition for checking method calls on the correct client

PR #70, opened 2025-03-04
Remove PHOTO and add FRAGILE to ItemCategory

anotheroneofthese:


MormonJesus69420 (approved):

MormonJesus69420:
Should probably check which categories can go into synq, cause I think some of them would not go in there

PR #73, opened 2025-03-14
MLT-121 Fix issues with order IDs from SynQ

MormonJesus69420:
Introduced normalization of order IDs to handle prefixed IDs consistently. Updated relevant logic in the SynqController and corresponding tests to use the normalized order IDs. Added a utility function `normalizeOrderId` and addressed potential edge cases in its implementation.


anotheroneofthese (approved):
Interesting delimiter choice

PR #74, opened 2025-03-18
MLT-0119 Location and Category rulesets

anotheroneofthese:
Implements handling of location and item categories when sending orders to storage systems.


src/main/kotlin/no/nb/mlt/wls/domain/WLSService.kt
```diff
@@ -164,7 +164,7 @@ class WLSService(
                 outboxRepository.save(OrderDeleted(order.hostName, order.hostOrderId))
             } ?: throw RuntimeException("Could not delete order")
 
-        outboxMessageProcessor.handleEvent(outBoxMessage)
+        processMessageAsync(outBoxMessage)
```
MormonJesus69420, line 167: Gonna need fixing after my PR

src/main/kotlin/no/nb/mlt/wls/domain/ports/outbound/ItemRepository.kt
```diff
@@ -25,6 +25,7 @@ interface ItemRepository {
         location: String
     ): Item
 
+    // TODO - This is MongoDB specific code, and should be moved into infrastructure instead
```
MormonJesus69420, line 28: I don't see how this is Mongo specific?

src/main/kotlin/no/nb/mlt/wls/infrastructure/synq/SynqAutostoreAdapter.kt
```diff
@@ -0,0 +1,119 @@
...
+    override suspend fun canHandleLocation(location: String): Boolean {
+        return when (location.uppercase()) {
+            "SYNQ_AUTOSTORE" -> true
+            "AUTOSTORE_WAREHOUSE" -> true
+            else -> false
+        }
+    }
+
+    override fun canHandleItem(item: Item): Boolean {
+        // SynQ can handle both NONE and FREEZE environments, so this is not checked
```
MormonJesus69420, line 110: I don't think Autostore can handle freeze though

src/main/kotlin/no/nb/mlt/wls/infrastructure/synq/SynqOrderPayload.kt
```diff
@@ -32,13 +33,18 @@ data class SynqOrderPayload(
...
+        AUTOSTORE("Autostore");
 
         @JsonValue
         override fun toString(): String {
             return type
         }
     }
+
+    companion object {
+        const val DELIMITER = "---"
```
MormonJesus69420, line 46: I think the delimiter we use will be the same regardless of what storage we send data to. So it might be best to move it to storage facade?

src/main/kotlin/no/nb/mlt/wls/infrastructure/synq/SynqOrderPayload.kt
```diff
@@ -67,10 +73,18 @@ data class ShippingAddress(
...
 }
 
+fun Order.toAutostorePayload(): SynqOrderPayload {
+    val payload = this.toSynqPayload()
+    return payload.copy(
+        orderId = hostName.toString().uppercase() + "-AS" + DELIMITER + hostOrderId,
+        orderType = SynqOrderPayload.SynqOrderType.AUTOSTORE
+    )
+}
+
```
MormonJesus69420, line 83: Isn't it better to just add a parameter to the `toSynqPayload` function?
Also I think we should be consistent and append either `AS` or `SD` to all orders

src/main/kotlin/no/nb/mlt/wls/infrastructure/synq/SynqAutostoreAdapter.kt
```diff
@@ -0,0 +1,119 @@
...
+                }
+            }
+            .onErrorMap(WebClientResponseException::class.java) { createServerError(it) }
+            .onErrorComplete(SynqError.DuplicateItemException::class.java)
+            .awaitSingle()
+    }
+
+    override suspend fun createOrder(order: Order) {
+        // Wrap the order in the way SynQ likes it
+        val orders = SynqOrder(listOf(order.toSynqPayload()))
```
MormonJesus69420, line 50: Shouldn't it be autostore payload?

MormonJesus69420 (changes requested):

src/main/kotlin/no/nb/mlt/wls/domain/WLSService.kt
anotheroneofthese: can't wait

src/main/kotlin/no/nb/mlt/wls/domain/ports/outbound/ItemRepository.kt
anotheroneofthese: As discussed this is not exactly Mongo-specific code, but it is somewhat specific to repositories. As-is it is fine, although it might be ideal to move it where it truly is required, which is to be used as a composite key for databases

src/main/kotlin/no/nb/mlt/wls/infrastructure/synq/SynqOrderPayload.kt
```diff
@@ -107,12 +117,14 @@ fun Packaging.toSynqPackaging(): SynqPackaging =
...
-        Order.Type.LOAN -> SynqOrderPayload.SynqOrderType.STANDARD
-        Order.Type.DIGITIZATION -> SynqOrderPayload.SynqOrderType.STANDARD
-    }
+// this needs a better name
+fun generatePostfix(type: SynqOrderPayload.SynqOrderType): String {
+    return "-" +
+        when (type) {
+            SynqOrderPayload.SynqOrderType.AUTOSTORE -> "AS"
+            else -> "SD"
+        }
```
MormonJesus69420, line 126:
```suggestion
    return
        when (type) {
            SynqOrderPayload.SynqOrderType.AUTOSTORE -> "-AS"
            else -> "-SD"
        }
```

MormonJesus69420 (approved):

src/main/kotlin/no/nb/mlt/wls/infrastructure/synq/SynqAutostoreAdapter.kt
```diff
@@ -0,0 +1,123 @@
...
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.stereotype.Component
+import org.springframework.web.reactive.function.client.WebClient
+import org.springframework.web.reactive.function.client.WebClientResponseException
+import reactor.core.publisher.Mono
+import java.net.URI
+
+// TODO - There is room to deduplicate the calls in this adapter
+@Component
+class SynqAutostoreAdapter(
```
MormonJesus69420, line 22: I know I am being mega nitpicky right now, and it does not need to be done now, but it might be good to change `SynqAdapter` name to something like `SynqStandardAdapter` for added clarity. I.E. This is adapter for Autostore SynQ, the other is for Standard SynQ.

src/main/kotlin/no/nb/mlt/wls/infrastructure/synq/SynqOrderPayload.kt
```diff
@@ -67,10 +69,14 @@ data class ShippingAddress(
...
+
+fun Order.toSynqPayload(type: SynqOrderPayload.SynqOrderType) =
     SynqOrderPayload(
-        orderId = hostName.toString().uppercase() + "---" + hostOrderId,
-        orderType = orderType.toSynqOrderType(),
+        orderId = hostName.toString().uppercase() + generatePostfix(type) + DELIMITER + hostOrderId,
+        orderType = type,
         // When order should be dispatched, AFAIK it's not used by us as we don't receive orders in future
         dispatchDate = LocalDateTime.now(),
         // When order was made in SynQ, if we want to we can omit it and SynQ will set it to current date itself
```
MormonJesus69420, line 82: Dunno what others think, but couldn't this whole thing be extracted to a function instead of just using the `generatePostfix`, so make a function called `generatePrefix` which combines both hostname, storage type, and delimited into one. Would make this a bit cleaner

MormonJesus69420: @tomolse?

src/main/kotlin/no/nb/mlt/wls/domain/ports/outbound/ItemRepository.kt
MormonJesus69420: I can see that, we had good arguments for and against it. I think we can leave it as is for now.

src/main/kotlin/no/nb/mlt/wls/application/synqapi/synq/SynqController.kt
```diff
@@ -251,7 +252,7 @@ private fun mapHostNameString(hostNameString: String?): HostName? {
 }
 
 private fun normalizeOrderId(orderId: String): String {
-    val orderIdWithoutPrefix = orderId.substringAfter("---", orderId)
+    val orderIdWithoutPrefix = orderId.substringAfter(StorageSystemFacade.DELIMITER, orderId)
```
MormonJesus69420, line 255: Also, not 100 sure, but maybe might want to move this to the SynqOrderPayload, so we have functions for both adding the prefix, and removing it in one place

MormonJesus69420 (approved):

PR #79, opened 2025-03-28
MLT-0126 Order Fixes

anotheroneofthese:
Fixes for bugs that prevent sending orders to SynQ


MormonJesus69420 (approved):

src/main/kotlin/no/nb/mlt/wls/application/hostapi/ExceptionHandler.kt
```diff
@@ -27,6 +28,13 @@ class ExceptionHandler {
             .body(ErrorMessage(message))
     }
 
+    @ExceptionHandler(DecodingException::class)
+    fun handleDecodingException(e: DecodingException): ResponseEntity<ErrorMessage> {
+        return ResponseEntity
+            .badRequest()
+            .body(ErrorMessage(e.message ?: "Error decoding request body. Likely missing a field."))
```
MormonJesus69420, line 35: maybe add the body to the message?

anotheroneofthese: This is as good as it gets with this exception unfortunately

PR #81, opened 2025-04-02
MLT-0128 Validation Rework

anotheroneofthese:
Reworks validation for data objects to use Jakarta. Refactors the test to support this as well

### DIFF
--- a/src/main/kotlin/no/nb/mlt/wls/application/synqapi/synq/SynqOrderPickingConfirmationPayload.kt
+++ b/src/main/kotlin/no/nb/mlt/wls/application/synqapi/synq/SynqOrderPickingConfirmationPayload.kt
@@ -167,8 +167,9 @@ data class OrderLine(
 ) {
     @Throws(ValidationException::class)
     fun validate() {
-        // TODO - migrate this to changes done in MLT-133 when merged in
-        if (HostName.entries.toTypedArray().none { it.name.uppercase() == hostName.uppercase() }) {
+        try {
+            HostName.fromString(hostName)
+        } catch (e: IllegalArgumentException) {
             throw ValidationException("Order Line's host name: '$hostName' is not valid")
         }
     }
--- a/src/main/kotlin/no/nb/mlt/wls/application/synqapi/synq/SynqOrderPickingConfirmationPayload.kt
+++ b/src/main/kotlin/no/nb/mlt/wls/application/synqapi/synq/SynqOrderPickingConfirmationPayload.kt
@@ -6,6 +6,7 @@ import jakarta.validation.constraints.Min
 import jakarta.validation.constraints.NotBlank
 import jakarta.validation.constraints.NotEmpty
 import jakarta.validation.constraints.PositiveOrZero
+import no.nb.mlt.wls.domain.ValidHostName
 import no.nb.mlt.wls.domain.model.HostName
 import no.nb.mlt.wls.domain.ports.inbound.ValidationException
 
@@ -56,12 +57,6 @@ data class SynqOrderPickingConfirmationPayload(
     @field:NotBlank(message = "Picking update's warehouse can not be blank")
     val warehouse: String
 ) {
-    @Throws(ValidationException::class)
-    fun validate() {
-        // Validates the hostname based on the value of hostname in order lines
-        getValidHostName()
-    }
-
     @Throws(ValidationException::class)
     fun getValidHostName(): HostName {
         val hostName = getHostNameString()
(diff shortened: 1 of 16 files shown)


src/main/kotlin/no/nb/mlt/wls/domain/NullableNotBlank.kt
```diff
@@ -0,0 +1,32 @@
...
+
+class NullableNotBlankImpl : ConstraintValidator<NullableNotBlank, String> {
+    override fun isValid(
+        value: String?,
+        context: ConstraintValidatorContext?
+    ): Boolean {
+        if (value == null) {
+            return true
+        }
+        return value.isNotBlank()
```
MormonJesus69420, line 30: Mayhaps?
```suggestion
        return value == null || value.isNotBlank()
```

src/main/kotlin/no/nb/mlt/wls/domain/model/Order.kt
```diff
@@ -188,38 +189,21 @@ data class Order(
...
         val addressLine1: String?,
+        @field:NullableNotBlank(message = "Invalid address: address line must not be blank")
         val addressLine2: String?,
+        @field:NullableNotBlank(message = "Invalid address: postcode must not be blank")
         val postcode: String?,
+        @field:NullableNotBlank(message = "Invalid address: city must not be blank")
         val city: String?,
+        @field:NullableNotBlank(message = "Invalid address: region must not be blank")
         val region: String?,
+        @field:NullableNotBlank(message = "Invalid address: country must not be blank")
```
MormonJesus69420, line 204:
```suggestion
        @field:NullableNotBlank(message = "Invalid address: recipient must not be blank if defined")
        val recipient: String?,
        @field:NullableNotBlank(message = "Invalid address: address line must not be blank if defined")
        val addressLine1: String?,
        @field:NullableNotBlank(message = "Invalid address: address line must not be blank if defined")
        val addressLine2: String?,
        @field:NullableNotBlank(message = "Invalid address: postcode must not be blank if defined")
        val postcode: String?,
        @field:NullableNotBlank(message = "Invalid address: city must not be blank if defined")
        val city: String?,
        @field:NullableNotBlank(message = "Invalid address: region must not be blank if defined")
        val region: String?,
        @field:NullableNotBlank(message = "Invalid address: country must not be blank if defined")
```

src/main/kotlin/no/nb/mlt/wls/application/synqapi/synq/SynqOrderPickingConfirmationPayload.kt
```diff
@@ -35,35 +40,26 @@ data class SynqOrderPickingConfirmationPayload(
...
-
-        if (operator.isBlank()) {
-            throw ValidationException("Picking update's operator can not be blank")
-        }
-
-        if (warehouse.isBlank()) {
-            throw ValidationException("Picking update's warehouse can not be blank")
-        }
         // Validates the hostname based on the value of hostname in order lines
         getValidHostName()
```
MormonJesus69420, line 62: Is this needed if we validate HostName in order lines?

src/main/kotlin/no/nb/mlt/wls/application/synqapi/synq/SynqOrderPickingConfirmationPayload.kt
```diff
@@ -124,79 +120,57 @@ data class OrderLine(
...
 ) {
     @Throws(ValidationException::class)
     fun validate() {
-        if (hostName.isBlank()) {
-            throw ValidationException("Order Line's host name can not be blank")
-        }
-
-        if (HostName.entries.toTypedArray().none { it.name.uppercase() == hostName.uppercase() }) {
+        try {
+            HostName.fromString(hostName)
```
MormonJesus69420, line 171: Maybe remake that into a validator annotation, same as NullableNotBlank. That way we can get rid of validate functions on every model

MormonJesus69420 (changes requested):

src/test/kotlin/no/nb/mlt/wls/synq/model/SynqModelValidationTest.kt
```diff
@@ -10,132 +13,94 @@ import no.nb.mlt.wls.domain.ports.inbound.ValidationException
...
-        then(thrown)
-            .isNotNull()
-            .isInstanceOf(ValidationException::class.java)
-            .hasMessageContainingAll("host name", "blank")
+        assert(validator.validate(orderLine).isNotEmpty())
     }
 
     @Test
     fun `OrderLine with invalid hostName should fail validation`() {
+        // TODO - this check doesn't detect invalid hostnames
```
MormonJesus69420, line 82: pic

MormonJesus69420 (approved):
GG, NO RE

PR #83, opened 2025-04-09
MLT-127 Group events for processing

MormonJesus69420:
We noticed that events were processed sequentially, and if one failed, we aborted processing of all subsequent events. Neither of those are desired.

If one event fails, others should be attempted because they might succeed.

If we attempt other events when one fails, and we process things sequentially it might lead to issues.

For example, we have 3 events for order1, first event says order started, the second says order picked, the third says order completed.  If second event failed but third went through, on the next run we would overwrite status completed with status picked.

Hence, we group all events by ID (either order or item) and process them in separate batches/groups.  When one event in batch fails, we will abort processing other events in batch.

Thus, all events that can be processed get processed, and they get processed in chronological order.

---

I have also fixed other things:
- Simplify testing of EventProcessors --> MLT-125
- Events retrieved in order by their timestamp
- Added a run config to project so we can all run project with same settings.
- This should also fix the fragile tests in WLS Service caused by switch from data class to regular class for Item --> MLT-132

---

See Jira task for test data I've used and results


.run/Run Hermes.run.xml
```diff
@@ -1,6 +1,9 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="Run Hermes" type="SpringBootApplicationConfigurationType" factoryName="Spring Boot">
     <option name="ACTIVE_PROFILES" value="local-dev" />
+    <envs>
+      <env name="ORDER_HANDLER_EMAIL" value="daniel.salwerowicz@nb.no" />
```
anotheroneofthese, line 5: Should probably update this to use and send test emails with Mailhog locally, so that I don't have to pester you constantly :slightly_smiling_face:

src/main/kotlin/no/nb/mlt/wls/infrastructure/StorageEventProcessorAdapter.kt
```diff
@@ -109,17 +121,44 @@ class StorageEventProcessorAdapter(
...
+        val updatedOrder = event.updatedOrder
+        val items =
+            itemRepository.getItems(
+                updatedOrder.hostName,
+                updatedOrder.orderLine.map { it.hostId }
+            )
+
+        mapItemsOnLocation(items).forEach { (storageSystemFacade, itemList) ->
+            if (storageSystemFacade == null) {
+                logger.info { "SEO: Could not find a storage system to handle items: $itemList" }
```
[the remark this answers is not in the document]
MormonJesus69420: Yeah, either warning or error.

.run/Run Hermes.run.xml
MormonJesus69420: I could just remove it and everyone uses their own stuff as needed. But adding a mail server and whole shebang to docker compose sounds like it's own task.
I can just leave it here for now, and if you need to test it you can just change the email to your own.

anotheroneofthese (approved):

PR #94, opened 2025-04-24
MLT-0141 Add UpdateItem

anotheroneofthese:
Required for handling SynQ and AutoStore correctly

### DIFF
--- a/src/main/kotlin/no/nb/mlt/wls/application/synqapi/synq/SynqBatchMoveItemPayload.kt
+++ b/src/main/kotlin/no/nb/mlt/wls/application/synqapi/synq/SynqBatchMoveItemPayload.kt
@@ -142,6 +142,10 @@ data class Product(
     )
     @field:PositiveOrZero(message = "Quantity on hand must not be negative. It must be zero or higher")
     val quantityOnHand: Int?,
+    @Schema(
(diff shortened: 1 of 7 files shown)


src/main/kotlin/no/nb/mlt/wls/application/synqapi/synq/SynqBatchMoveItemPayload.kt
```diff
@@ -156,7 +165,36 @@ data class Product(
...
+        location: String
+    ): MoveItemPayload {
+        val quantity =
+            if (prevLocation == "AutoStore_Warehouse") {
+                quantityMove ?: throw ItemMovingException("Quantity moved must not be null")
+                quantityMove.unaryMinus()
+            } else {
+                quantityMove ?: throw ItemMovingException("Quantity moved must not be null")
+                quantityMove
+            }
```
MormonJesus69420, line 180:
```suggestion
        var quantity = quantityMove ?: throw ItemMovingException("Quantity moved must not be null")
        if (prevLocation == "AutoStore_Warehouse") {
           quantity = quantity.unaryMinus()
        }
```

MormonJesus69420 (approved):

PR #99, opened 2025-05-15
MLT-0079 Sending Orders to Kardex

anotheroneofthese:


src/main/kotlin/no/nb/mlt/wls/domain/ports/outbound/StorageSystemFacade.kt
```diff
@@ -28,6 +29,8 @@ interface StorageSystemFacade {
     suspend fun canHandleLocation(location: String): Boolean
 
     fun canHandleItem(item: Item): Boolean
+
+    fun supportsEvent(event: StorageEvent): Boolean
```
MormonJesus69420, line 33: First, might as well have a default implementation.
Second, should call it every place, not just in update order.
Third, omg look at you, you look so cool, new haircut?
Fourth, I think we should look at other ways to handle the fact that Kardex doesn't allow order editing (set status on order from get go so update will be rejected, delete old order & create new one, or something else) Might be good to get some input from @tomolse and others in team.

MormonJesus69420: Oh another option, delete order editing functionality completely, as for now nobody can edit any order in Kardex or SynQ

anotheroneofthese: We will have to review that functionality in general, considering the interactions with the storage systems. If we were to remove `updateOrder` then this entire implementation is unneeded.

PR #122, opened 2025-06-11
Add user and client for dashboard app to keycloak realm file

tomolse:


tomolse:
> Not sure what password you've set for the user, but for local testing, to keep things simple we just use the same password as username, so kardex client has password kardex, synq has synq, etc. Probably best to do the same for this one too if it was not done. otherwise looks ok.

Fixed

MormonJesus69420 (approved):

PR #126, opened 2025-06-25
MLT-0181 Fix WITH_LENDER being overridden during stock sync

anotheroneofthese:


MormonJesus69420 (approved):
Looks good, could've merged it with the previous if statement, but I don't think updating the quantity will hurt in this case.

PR #128, opened 2025-06-27
MLT-0183 Order Domain status updating

anotheroneofthese:


src/main/kotlin/no/nb/mlt/wls/domain/model/Order.kt
```diff
@@ -70,26 +74,34 @@ data class Order(
...
-    private fun isOrderClosed(): Boolean = listOf(Status.COMPLETED, Status.DELETED, Status.RETURNED).contains(status)
+    /**
+     * Order can not receive any further updates
+     */
+    private fun isOrderClosed(): Boolean = listOf(Status.DELETED, Status.RETURNED).contains(status)
+
+    /**
+     * Order is completed, but is not returned yet
+     */
+    private fun isOrderComplete(): Boolean = listOf(Status.COMPLETED, Status.DELETED).contains(status)
```
tomolse, line 104: Should rename this to `isOrderPicked()` and fix check to represent that.

tomolse (approved):

src/main/kotlin/no/nb/mlt/wls/domain/model/Order.kt
```diff
@@ -122,6 +134,38 @@ data class Order(
         }
     }
 
+    fun updateStatus(status: Status): Order {
+        if (isOrderClosed()) {
+            throw ValidationException("The order is already closed, and can therefore not be changed")
+        }
+        return when (this.status) {
```
MormonJesus69420, line 141: Dunno how kosher is that in Kotlin world, but couldn't we just math it out with enums? I.E. can't change enum to a lower value one. requires that enums values are in correct order, but I mean they already are.

```kotlin
    enum class Status {
        NOT_STARTED,
        IN_PROGRESS,
        COMPLETED,
        DELETED,
        RETURNED
    }
```

PR #133, opened 2025-07-01
Cleanup Order class and create tests

tomolse:


src/main/kotlin/no/nb/mlt/wls/domain/model/Order.kt
```diff
@@ -18,43 +17,32 @@ data class Order(
...
-    }
-
     private fun setOrderLineStatus(
         hostIds: List<String>,
         status: OrderItem.Status
     ): Order {
-        if (isOrderClosed()) {
-            throw IllegalOrderStateException("Order is already closed with status: $status")
+        if (isClosed()) {
+            throw ValidationException("Order is already closed with status: $status")
```
MormonJesus69420, line 25: Changing to ValidationException might have broken at least one test (the one that was meant to return 409 conflict error)

src/main/kotlin/no/nb/mlt/wls/domain/model/Order.kt
```diff
@@ -94,16 +82,16 @@ data class Order(
...
+     * Order cannot receive any further updates
      */
-    private fun isOrderClosed(): Boolean = listOf(Status.DELETED, Status.RETURNED).contains(status)
+    fun isClosed(): Boolean = listOf(Status.DELETED, Status.RETURNED).contains(status)
 
     /**
      * Order is picked and finished, but is not returned yet
      */
-    private fun isOrderPicked(): Boolean = listOf(Status.COMPLETED).contains(status)
+    private fun isPicked(): Boolean = listOf(Status.COMPLETED).contains(status)
```
MormonJesus69420, line 92: Why not?
```suggestion
    private fun isPicked(): Boolean = status == Status.COMPLETED
```

MormonJesus69420 (approved):

PR #137, opened 2025-07-07
MLT-178 GET endpoints

MormonJesus69420:
Add GET endpoints for Items and Orders, for use in dashboard app


MormonJesus69420:
Merging it on my own, good idea to have @anotheroneofthese review it when he's back from holidays.

PR #144, opened 2025-07-31
MLT-201 Properly test code inside TransactionPort

MormonJesus69420:
As we haven't actually been properly testing it, I think it's time to fix that. Every public method should be tested at least once for correctness inside the TransactionPort.

Doing a general cleanup too as I come over something that bothers me


src/main/kotlin/no/nb/mlt/wls/domain/model/Order.kt
```diff
@@ -101,12 +95,6 @@ data class Order(
 
     private fun createOrderAddress(): Address = Address(null, null, null, null, null, null, null)
 
-    /**
-     * Delete the order as long as it is possible.
-     *
-     * This validates if the order can be deleted, and sets the status on it if it can.
-     * Deleting it from the storage systems is handled by the outbox processor.
-     */
```
MormonJesus69420, line 109: We do a little gaslighting 👌🏼😩💯

src/main/kotlin/no/nb/mlt/wls/domain/WLSService.kt
```diff
@@ -248,11 +229,28 @@ class WLSService(
...
+
+        processCatalogEventAsync(catalogEvent)
+    }
+
     override suspend fun deleteOrder(
         hostName: HostName,
         hostOrderId: String
     ) {
-        val order = getOrderOrThrow(hostName, hostOrderId).deleteOrder()
+        val order = getOrderOrThrow(hostName, hostOrderId)
```
anotheroneofthese, line 253:
```suggestion
        val order = getOrderOrThrow(hostName, hostOrderId)
        val deletedOrder = order.deleteOrder()
```
The call to remove the order should not be removed. Could be made more explicit by putting it on its own line

MormonJesus69420: Why? It does nothing

MormonJesus69420: fixed

src/test/kotlin/no/nb/mlt/wls/domain/WLSServiceTest.kt
```diff
@@ -1,139 +1,184 @@
...
-import no.nb.mlt.wls.domain.ports.outbound.EventProcessor
-import no.nb.mlt.wls.domain.ports.outbound.EventRepository
-import no.nb.mlt.wls.domain.ports.outbound.ItemRepository
-import no.nb.mlt.wls.domain.ports.outbound.OrderRepository
-import no.nb.mlt.wls.domain.ports.outbound.StorageSystemException
-import no.nb.mlt.wls.domain.ports.outbound.StorageSystemFacade
-import no.nb.mlt.wls.domain.ports.outbound.TransactionPort
+import no.nb.mlt.wls.domain.ports.inbound.*
+import no.nb.mlt.wls.domain.ports.inbound.UpdateItem.UpdateItemPayload
+import no.nb.mlt.wls.domain.ports.outbound.*
```
anotheroneofthese, line 17: Wildcard imports are not allowed in our current linting

anotheroneofthese (changes requested):

anotheroneofthese (approved):

PR #164, opened 2025-08-14
Update dependencies

anotheroneofthese:
Supersedes #157, #158, #159, #160, #161, #162.

MockWebServer requires a noticeable refactor, and also some extra pins on dependencies managed by micrometer/Spring. Not sure if this is something we should include in this batch.


MormonJesus69420 (dismissed):
Looks good, just unsure if the MWS upgrade and refactor should be its own PR or no. :shipit:

PR #179, opened 2025-09-16
MLT-0216 Fix Swagger documentation descriptions

anotheroneofthese:
Use the GroupedOpenApi bean exclusively, allowing for descriptions per schema


src/main/kotlin/no/nb/mlt/wls/application/hostapi/config/SwaggerConfig.kt
```diff
@@ -27,42 +24,34 @@ import org.springframework.context.annotation.Primary
...
+                    Applications that want to use Hermes must authenticate with a JWT token.
+                    It is provided by the NLN's instance of Keycloak with help of a Service Account client.
+                    Please contact the MLT team for help getting set up with one or to reset credentials.
+                    """.trimIndent()
+                it.info.contact =
+                    Contact()
+                        .name("MLT team at the National Library of Norway")
+                        .email("mlt@nb.no")
+                        .url("https://www.nb.no")
+                it.info.version = "1.0.20"
```
MormonJesus69420, line 53: Might wanna think if we want to have this, and if so, might be good to have it automagically updated when we push new version?

PR #295, opened 2026-04-13
Prevent BIBLIOFIL items in Preservation SynQ

anotheroneofthese:
Minor fix


MormonJesus69420 (changes requested):

anotheroneofthese:
We do kinda need this temporarily, as it causes some headache in our testing environment

PR #297, opened 2026-04-21
MLT-0336 Fix confidential items not syncing correctly

anotheroneofthese:


anotheroneofthese:
We are at the point where we want metadata from hosts/catalogs to determine metadata, not the other way around

